PR #6, opened 2017-08-20
User sign up

akintolga:
- Initial sign up page for new users

### DIFF
--- a/newsroom/auth.py
+++ b/newsroom/auth.py
@@ -1,4 +1,10 @@
 import flask
+from newsroom.utils import query_resource
+from superdesk import get_resource_service
+from superdesk.utc import utcnow
+from datetime import timedelta
+from flask import current_app as app
+import bcrypt
 
 
 blueprint = flask.Blueprint('auth', __name__)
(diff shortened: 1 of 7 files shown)


akintolga:
@petrjasek This pr relies on https://github.com/superdesk/superdesk-core/pull/1013
I thought it would be better to keep auth, users and company logic in content-api
Let me know what you think

petrjasek:
I would rather have it here, eventually everything related to content api should go here imo, there won't be any use for it in superdesk

newsroom/auth.py
```diff
@@ -8,15 +14,95 @@
...
+            user = get_resource_service('users').find_one(req=None, _id=user['_id'])
+
+            flask.session['name'] = user.get('name')
             return flask.redirect(flask.url_for('news.index'))
-        kwargs['credentials_error'] = True
-        kwargs['username'] = flask.request.form.get('username', '')
-    return flask.render_template('auth_login.html', **kwargs)
+        except Exception as ex:
+            flask.flash('Invalid login: {} Please try again.'.format(str(ex)))
+            return flask.redirect(flask.url_for('auth.login'))
```
petrjasek: I'm not sure if we should redirect on error in general, imo it's better to render the page again and keep the data where appropriate (like email here).

newsroom/templates/auth_login.html
```diff
@@ -7,8 +7,18 @@
 <p>Invalid credentials.</p>
 {% endif %}
 
+{% with messages = get_flashed_messages() %}
```
petrjasek: imo this should go into layout so we don't repeat it

petrjasek:
would be good to use gettext for all strings

akintolga:
gettext is done.

newsroom/templates/signup.html
```diff
@@ -0,0 +1,46 @@
...
+        <option value="Sales">Sales Representative</option>
+        <option value="CTO">CTO</option>
+        <option value="CEO">CEO</option>
+    </select></p>
+
+    <label><b>Size of your company</b></label>
+    <p><select name="company_size">
+        <option value="0-10">0-10</option>
+        <option value="11-100">11-100</option>
+        <option value=">100">>100</option>
```
jerome-poisson: should not the **>** be escaped to **&amp;gt;**?

newsroom/auth.py
```diff
@@ -1,4 +1,11 @@
 import flask
+from newsroom.utils import query_resource
+from superdesk import get_resource_service
+from superdesk.utc import utcnow
+from datetime import timedelta
+from flask import current_app as app
+import bcrypt
+from flask_babel import gettext
```
jerome-poisson: gettext is usually imported as **_** to make it more readable

PR #7, opened 2017-08-27
User list and companies

akintolga:
Major changes in this PR:
- User sign up
- User list
- Edit a user
- Create a company
- Company list
- Edit a company
- JQuery setup
- Template layout improvements
- Test settings and new unit tests


newsroom/users/tests.py
```diff
@@ -0,0 +1,30 @@
+from tests import TestCase
+from flask import url_for
+
+
+class UsersTestCases(TestCase):
```
petrjasek: would rather put tests into `tests` folder so when this will be installable via pip it won't get them

newsroom/auth/tests.py
```diff
@@ -0,0 +1,40 @@
+from tests import TestCase
+from flask import url_for
+
+
+class AuthTestCases(TestCase):
```
petrjasek: this should be also in `tests/auth_tests.py` imo

petrjasek:
in general I would avoid that javascript and jquery in particular :wink:
imo would be easier to stick with html forms, or is there any benefit?

newsroom/users/tests.py
akintolga: No problem, I'll move them to test folder.

akintolga:
I thought we were going to have side panel as we had in superdesk for users (I vaguely remember Fritz suggesting following superdesk style in settings), that's why I included jquery. If we want to use a panel to edit (or preview) we need some javascript.

I don't see any problem using forms only, it should be fairly easy to switch to forms. I'll leave it like this if it is ok for this PR. I can change it in later PRs.

petrjasek:
it's ok for this PR, but I would rather avoid writing javascript for every form, if that means ditching the sidebar UI lets do this.. what do you think @hlmnrmr ?

Issue #8, opened 2017-09-07
Integrate pytest instead of nosetests


naspeh:
pytest is modern tool for testing in Python world. It is active, the latest version from 2017-08-09.

The latest version of `nose`: https://pypi.python.org/pypi/nose/1.3.7 from 2015-06-02
There is also `nose2` (same author) https://pypi.python.org/pypi/nose2/0.6.5 from 2016-06-29

Docs: https://docs.pytest.org/en/latest/
For BDD: https://pypi.python.org/pypi/pytest-bdd

`newsroom` project has a little tests yet, so it'll be easy to transform it, I can help with this step.

@petrjasek, @akintolga what do you think?

petrjasek:
looks good, be my guest :wink:

akintolga:
Sounds good to me to 👍

naspeh:
Great, I'm going to integrate it today

naspeh:
Ready in #9 :)

PR #9, opened 2017-09-08
Integrate pytest instead of nosetests

naspeh:
According to the proposal in #8

Some usage
```bash
# run only tests from this module
pytest tests/test_newsroom.py

# run tests which match this pattern (in this case only one test)
pytest -k new_user_signup_succeeds

# coverage
pytest --cov=newsroom
```


naspeh:
`pytest` is green on Travis

`npm run lint` fails

akintolga:
Merging this now and I'll have another PR for js lint problems

PR #11, opened 2017-09-10
[SDAN-52] - Validate the email of signed up user

akintolga:
- An email will be sent for new users to validate email address
- Users can ask for validation email again
- Login will fail for not-yet-validated users
- Login will fail if user's company is disabled [SDAN-49]
- Administrators can resend a validation email again [SDAN-43]
- Administrators can send a reset password token [SDAN-42]

### DIFF
--- a/newsroom/auth/views.py
+++ b/newsroom/auth/views.py
@@ -153,11 +153,18 @@ def token():
     token_type = flask.request.args['token_type']
     if form.validate_on_submit():
         user = get_resource_service('users').find_one(req=None, email=form.email.data)
-        if user is not None:
-            updates = {}
-            _add_token_data(updates)
-            get_resource_service('users').patch(id=ObjectId(user['_id']), updates=updates)
-            send_validate_account_email(user['name'], user['email'], updates['token'])
+        send_token(user, token_type)
         flask.flash('A new validation token has been sent. Please check your emails')
         return flask.redirect(flask.url_for('auth.login'))
     return flask.render_template('request_token.html', form=form, token_type=token_type)
+
+
+def send_token(user, token_type='validate'):
+    if user is not None and not user.get('is_validated'):
+        updates = {}
+        _add_token_data(updates)
+        get_resource_service('users').patch(id=ObjectId(user['_id']), updates=updates)
+        if token_type == 'validate':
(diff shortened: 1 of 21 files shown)


akintolga:
Depends on https://github.com/superdesk/superdesk-core/pull/1027

newsroom/auth/views.py
```diff
@@ -15,7 +18,16 @@ def login():
         user = get_resource_service('auth_user').find_one(req=None, email=form.email.data)
         if user is not None and _is_password_valid(form.password.data.encode('UTF-8'), user):
             user = get_resource_service('users').find_one(req=None, _id=user['_id'])
-            if is_account_valid(user):
+
+            if not user.get('is_validated'):
+                flask.flash('Your email address needs validation.', 'danger')
```
petrjasek: these flash messages can use gettext

newsroom/auth/views.py
```diff
@@ -87,3 +116,61 @@ def _modify_user_data(new_user):
...
+
+def _add_token_data(user):
+    user['token'] = get_random_string()
+    user['token_expiry_date'] = utcnow() + timedelta(days=app.config['VALIDATE_ACCOUNT_TOKEN_TIME_TO_LIVE'])
+
+
+@blueprint.route('/validate')
+def validate_account():
+    token = flask.request.args['token']
+    if token:
```
petrjasek: maybe this route should be `/validate/<token>` so there is no need for this `if token:`

newsroom/email.py
```diff
@@ -0,0 +1,66 @@
...
+    :param subject: Subject text
+    :param text_body: Text Body
+    :param html_body: Html Body
+    :return:
+    """
+    msg = Message(subject=subject, sender=current_app.config['MAIL_DEFAULT_SENDER'], recipients=to)
+    msg.body = text_body
+    msg.html = html_body
+    app = current_app._get_current_object()
+    thr = Thread(target=send_async_email, args=[app, msg])
```
petrjasek: imo this is premature optimization :wink:

akintolga:
I think for the templates we may need to route to a different folder or filename. So if the local language is 'es' then we could have `/es/account_created.html` or `account_created.html.es` which would make easier to maintain.

PR #12, opened 2017-09-12
render list of items via react

petrjasek:
initial setup for react/redux app for wire

what do you think @akintolga ? seems like we will have to go with dynamic views, would create an app if needed for each

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,7 +6,7 @@ module.exports = {
     "globals": {
         "$": true
     },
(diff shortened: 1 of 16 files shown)


akintolga:
👍 I think this is a good move for wire items as the most interaction will be there. 

Do you think should I move settings pages to react as well or keep them with jquery for the time being?  Let me know what you think.

petrjasek:
given the current designs I think we will have to do it eventually, looks pretty interactive

PR #14, opened 2017-09-13
User creation by admins and password change

akintolga:
- [SDAN-58] - Creating users by administrators
- [SDAN-53] - User gets password reset email with a token when account is created or user clicks forgot password on login form. User then will land on reset password form with or without old password input field.
- [SDAN-44] - User deletion by administrators
- [SDAN-57] - Company deletion by administrators. This will also delete the users of the deleted company.


akintolga:
Depends on https://github.com/superdesk/superdesk-core/pull/1031

akintolga:
@petrjasek This is the last piece of -mostly server side- changes we can merge before refactoring settings area. I have now started wiring react to the settings page according to the new designs.

newsroom/auth/forms.py
```diff
@@ -41,3 +41,10 @@ class LoginForm(FlaskForm):
 
 class TokenForm(FlaskForm):
     email = StringField('Email', validators=[DataRequired(), Length(1, 64), Email()])
+
+
+class ResetPasswordForm(FlaskForm):
+    old_password = PasswordField('Old password', validators=[])
```
petrjasek: imo there should be no old password in reset process, that's the reason why you do it usually

akintolga: I see. I guess I was using ResetPassword form for both resetting (by a token) and for changing the password from user's My Details pages. Maybe it would be better if I separate them.

akintolga: Merging this for now but I'll revisit this area later on again.

PR #15, opened 2017-09-15
feat(wire): wire search

petrjasek:
fetch content from server using javascript instead or reloading
full page on search. supports browser history.

SDAN-14


petrjasek:
added karma for javascript testing

petrjasek:
breaking change :warning: use `npm run start` for dev server, seems to be the standard

PR #28, opened 2017-09-26
feat(wire): add copy action to item preview

petrjasek:


petrjasek:
@akintolga fyi I've made a wrapper for alertify so we can swap it if needed

PR #32, opened 2017-10-03
First name and Last name fields are created for users

akintolga:
It has been suggested to use `first_name` and `last_name` fields instead of a single `name` for compatibility to external systems (such as salesforce)


akintolga:
Depends on https://github.com/superdesk/superdesk-core/pull/1045

PR #34, opened 2017-10-03
implement wire item print action

petrjasek:
- using a specific print layout and template

SDAN-21

### DIFF
--- a/newsroom/templates/wire_item.html
+++ b/newsroom/templates/wire_item.html
@@ -1,0 +1,35 @@
+{% extends "layout.html" %}
+
+{% block title %}{{ item.headline }}{% endblock %}
+
+{% block content %}
+<article>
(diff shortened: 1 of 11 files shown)


newsroom/templates/wire_item.html
```diff
@@ -0,0 +1,35 @@
...
+            </div>
+
+            {{ item.body_html | safe }}
+        </div>
+        <div class="col">
+            <h5>{{ _('Metadata') }}</h3>
+            <dl>
+                {% if item.subject %}
+                <dt>{{ _('Category') }}</dt>
+                <dd>{{ item.subject|join(', ', attribute='name')}}</dd>
```
akintolga: I can't remember 100% but Category may correspond to `service` field

petrjasek: ok will check it later with karel

PR #35, opened 2017-10-04
[SDAN-56] - Display users of a company

akintolga:
- Users are displayed as a list on its own tab in company details
- There's no action defined for those users yet

### DIFF
--- a/assets/settings/actions.js
+++ b/assets/settings/actions.js
@@ -4,6 +4,13 @@ import server from 'server';
 
 export const SELECT_ITEM = 'SELECT_ITEM';
 export function selectItem(id) {
+    return function (dispatch) {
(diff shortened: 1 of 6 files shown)


petrjasek:
I've just made an extra endpoint for this - https://github.com/superdesk/newsroom/pull/36/files#diff-d7f91022988b2fbc0bc91c868afeeac6R75 - not sure if it would make sense for you to use this, it's more for users to get their peers, want to remove most of the fields eventually via projections and keep only first/last name

akintolga:
👍 
That's actually one end point I was considering but ended up modifying the `users/search` instead. I'll modify my PR to use that one. Thanks for that.

akintolga:
I have updated my changes to use new endpoint...

PR #40, opened 2017-10-05
introduce search sidebar with followed topics

petrjasek:
SDAN-9

### DIFF
--- a/assets/wire/components/FollowTopicModal.jsx
+++ b/assets/wire/components/FollowTopicModal.jsx
@@ -1,17 +1,23 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import Modal from 'components/Modal';
 import { connect } from 'react-redux';
 import { gettext } from 'utils';
 import { closeModal, submitFollowTopic } from 'wire/actions';
+
+import Modal, { ModalPrimaryButton, ModalSecondaryButton } from 'components/Modal';
+import CloseButton from 'components/CloseButton';
+import TextInput from 'components/TextInput';
+import CheckboxInput from 'components/CheckboxInput';
+
+const TOPIC_NAME_MAXLENGTH = 30;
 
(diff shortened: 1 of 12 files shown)


assets/wire/components/FollowTopicModal.jsx
```diff
@@ -41,23 +49,37 @@ class FollowTopicModal extends React.Component {
...
                             id="description"
-                            value={this.state.description}
-                            onChange={this.onChangeHandler('description')}
+                            value={this.state.query}
+                            readOnly={true}
                         />
                     </div>
+                    <div className="form-check">
+                        <label className="form-check-label">
+                            <input type="checkbox"
```
akintolga: Optionally for checkboxes in the form you could use this component:
https://github.com/superdesk/newsroom/blob/master/assets/components/CheckboxInput.jsx

assets/wire/components/FollowTopicModal.jsx
```diff
@@ -41,23 +49,37 @@ class FollowTopicModal extends React.Component {
...
                             value={this.state.label}
                             onChange={this.onChangeHandler('label')}
+                            maxLength={TOPIC_NAME_MAXLENGTH}
                         />
                     </div>
                     <div className="form-group">
-                        <label htmlFor="description">{gettext('Description')}</label>
-                        <textarea className="form-control"
+                        <label htmlFor="description">{gettext('Query')}</label>
+                        <input type="text"
```
akintolga: Fyi, there's an text input component for forms:
https://github.com/superdesk/newsroom/blob/master/assets/components/TextInput.jsx

It is the same thing but has a bit more in validation.

petrjasek:
@akintolga made use of those components with some modifications, we should probably create it for all such input types eventually, makes things consistent :+1:

PR #42, opened 2017-10-07
User sign up and login improvements

akintolga:
- [x] [SDAN-90] - User sign up sends an email rather than account creation
- [x] Introduced basic limiting for auth endpoints
- [x] [SDAN-48] - Handling of wrong passwords
- [x] Introduced Google reCaptcha for sign up page
- [x] [SDAN-50] - Remember me option for logins

### DIFF
--- a/newsroom/auth/forms.py
+++ b/newsroom/auth/forms.py
@@ -1,6 +1,5 @@
-from flask_wtf import FlaskForm
-from wtforms import StringField, PasswordField
-from wtforms import SelectField
+from flask_wtf import FlaskForm, RecaptchaField
+from wtforms import StringField, PasswordField, SelectField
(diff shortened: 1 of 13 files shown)


akintolga:
Hi @petrjasek here's few further notes about the changes:

**Rate Limiting:**  I used flask-limiter and set only on `auth` blueprint to use limiting. For now only limits the requests coming from the same ip address. We can also use redis down the track with moving window. 

**Caching:** I used flask-cache and configure it to use in-mem cache for now. Again we can use with redis at some later stage.

**ReCaptcha:** I used the V2 version where user clicks the checkbox for sign up page only. I created the secrets from google for domain `https://nr-master.test.superdesk.org` and I'll pass you those ones on slack. The keys should also make recaptcha functioning on localhost.

PR #49, opened 2017-10-11
[SDAN-87] - Action list on individual items

akintolga:
- An initial three-dot button on every item with the basic list of actions.


assets/components/ActionList.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+
+function ActionList({item, actions, onClose}) {
+    return (
+        <div className="wire-articles__item--list__actions">
+            <button
```
petrjasek: fyi there is CloseButton component

PR #55, opened 2017-10-14
UI Wiring

akintolga:
This includes the UI changes for 
- item cards [SDAN-92]
- item preview [SDAN-93]
- item search bar [SDAN-94]
- sidebar and previous versions

I think the best way to check this PR is to take a look at [this pr's build](http://nrpr-55.test.superdesk.org)

### DIFF
--- a/assets/index.scss
+++ b/assets/index.scss
@@ -343,6 +343,7 @@ article.list {
     font-size: 0.8rem;
     color: #707070;
     padding: 7px 20px;
+    margin-right: 5px;
     text-transform: uppercase;
     border: 1px solid #bababa;
     border-radius: 3px;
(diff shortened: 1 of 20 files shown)


assets/index.scss
```diff
@@ -400,7 +401,7 @@ article.list {
 }
 
 .wire-column__preview--open {
-    width: 440px;
+    width: 480px;
```
akintolga: Hi @aleksandarjelicic, I increased the preview panel size to 480 as the date and the action buttons were not fitting properly. Feel free to change it if it creates an issue.

akintolga:
Hi @darconny thanks for the notification. There was a stray div left somewhere in the markup. Now it should look better.

PR #62, opened 2017-10-18
update content notifications

petrjasek:


petrjasek:
will need to merge this first, then change deployment scripts and after that will start working on client

PR #64, opened 2017-10-19
[SDAN-98] - New fields for company

akintolga:
Also fixes [SDAN-99]


akintolga:
@petrjasek I'm bringing company end user resources back into newsroom, let me know if you see any issues with it

PR #68, opened 2017-10-19
 Adjusting news list css with implementation, wire item detail view, added breakpoints

aleksandarjelicic:
@akintolga I worked with sample data on actual template for wire item detail.
Most of the styles are done with modifier content content--item-detail, and if that's not possible for some reason please let me know, I will add modifier class somewhere else.

### DIFF
--- a/newsroom/templates/wire_item.html
+++ b/newsroom/templates/wire_item.html
@@ -3,33 +3,153 @@
 {% block title %}{{ item.headline }}{% endblock %}
 
 {% block content %}
-<article>
-    <pre>{{ item.slugline }}</pre>
+<div class="content content--item-detail">
 
-    <h1>{{ item.headline }}</h1>
-    <p>{{ _('By:') }} <b>{{ item.byline }}</b> {{ item.versioncreated|datetime_long }}</p>
+    <section class="content-header">
+        <div class="content-bar navbar justify-content-between">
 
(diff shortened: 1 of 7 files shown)


newsroom/templates/wire_item.html
```diff
@@ -3,33 +3,153 @@
...
+            </figure>
+
+            <div class="wire-column__preview__content--item-detail-text-wrap">
+                <div class="wire-column__preview__content--item-detail-item-text">
+                    <span class="wire-column__preview__slug">{{ item.slugline }}</span>
+                    <h2 class="wire-column__preview__headline">{{ item.headline }}</h2>
+                    <p class="wire-column__preview__author">{{ _('By:') }} <span class="bold">{{ item.byline }}</span> {{ item.versioncreated|datetime_long }}</p>
+                    <p class="wire-column__preview__lead">{{ item.description_html | safe }}</p>
+                    <div class="wire-column__preview__text">
+                        <p>{{ item.body_html | safe }}</p>
```
petrjasek: there should be no `<p>` tag around body, it usually has `<p>` tags or `<divs>`

aleksandarjelicic: Fixed

petrjasek:
shouldn't be item.headline in h1?

PR #71, opened 2017-10-21
[SDAN-106] - Wiring UI for single item details

akintolga:
- Opens the wire story on its own so it doesn't use `wire_item` template anymore
- You can preview the changes [here ](http://nrpr-71.test.superdesk.org/)

### DIFF
--- a/assets/utils.js
+++ b/assets/utils.js
@@ -102,7 +102,7 @@ export function fullDate(dateString) {
  */
 export function formatTime(dateString) {
     const date = parseDate(dateString);
-    return date.toLocaleTimeString();
+    return date.toLocaleTimeString([], {hour12: false, hour: '2-digit', minute:'2-digit'});
 }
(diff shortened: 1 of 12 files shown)


petrjasek:
I was wondering about this as well, question is how we would use this if user opens an item from email. or would you keep the other view (wire_item.html) as well?

aleksandarjelicic:
@darconny it's fine now, I made PR with few css fixes

akintolga:
@petrjasek I agree that we should be able to open a single story if the url is given. I was thinking about using internal routing (react-routing) to make the the url is the `source-of-truth` for preview and open actions.

I think the issue with the template might be the logic for action buttons. But if you think we should maintain the template we can look in to that a bit more.

petrjasek:
yep lets try the routing then, we can always fallback to the flask view

PR #72, opened 2017-10-22
User's my profile pages

akintolga:
- Initial wiring for the `followed topics`:

https://wiki.sourcefabric.org/display/AAPNEWS/UI+Design?preview=/65044524/77824022/AAP_Newsroom_Newswire_Follow_topic-6.png

### DIFF
--- a/assets/components/FollowTopicModal.jsx
+++ b/assets/components/FollowTopicModal.jsx
@@ -1,6 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
 import { gettext } from 'utils';
+import { closeModal, submitFollowTopic as submitWireFollowTopic } from 'wire/actions';
+import { submitFollowTopic as submitProfileFollowTopic } from 'user-profile/actions';
 
 import Modal, { ModalPrimaryButton, ModalSecondaryButton } from './Modal';
 import CloseButton from './CloseButton';
@@ -22,7 +25,7 @@ class FollowTopicModal extends React.Component {
     onSubmit(event) {
         event.preventDefault();
         if (this.state.topic.label) {
-            this.props.submit(this.state.topic);
+            this.props.submit(this.isNewTopic(), this.state.topic);
         }
     }
 
@@ -36,6 +39,10 @@ class FollowTopicModal extends React.Component {
     toggleNotifications() {
         const topic = Object.assign(this.state.topic, {notifications: !this.state.topic.notifications});
         this.setState({topic});
+    }
+
+    isNewTopic() {
+        return this.state.topic && !this.state.topic._id;
     }
 
     render() {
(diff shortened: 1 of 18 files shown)


assets/wire/components/WireApp.jsx
```diff
@@ -48,7 +50,13 @@ class WireApp extends React.Component {
     renderModal(specs) {
         if (specs) {
             const Modal = modals[specs.modal];
-            return <Modal key="modal" data={specs.data} />;
+            return <Modal
+                key="modal"
+                data={specs.data}
+                closeModal={this.props.closeModal}
+                submit={this.props.submitModal}
```
petrjasek: imo this part is confusing, any modal would get this while it only works for follow topic modal, thus I think it would be better to make that component smart and let it pick the action, how we avoid the duplication in actions/reducers is another problem

assets/user-profile/components/UserProfileMenu.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { gettext } from 'utils';
+
+
+function UserProfileMenu({isProfile, isFollowedTopics, isDownloadHistory, onClick}) {
```
petrjasek: not sure if we should create single app for all these, any benefits?

akintolga: I can think of few marginal benefits like some re-use of actions and reducers, no need for server round trip for every menu click, Maybe I can advocate for functionality for `user profile` feels like one app. But I can't create a strong case for single app to be honest.

assets/wire/components/WireApp.jsx
akintolga: I see. Let me think about this component.

assets/user-profile/components/UserProfileMenu.jsx
petrjasek: lets keep it for now, there is still some slight chance we would need to make it a modal, and then it would be easier

akintolga:
@petrjasek I have modified the `FollowTopicModal` so which action to be called will be decided within the component. Let me know if it makes sense.

assets/components/FollowTopicModal.jsx
```diff
@@ -84,13 +88,13 @@ FollowTopicModal.propTypes = {
...
     data: PropTypes.shape({
-        topic: PropTypes.string.isRequired,
+        topic: PropTypes.object.isRequired,
     }),
 };
 
 const mapDispatchToProps = (dispatch) => ({
-    closeModal: () => dispatch(closeModal()),
-    submit: (data) => dispatch(submitFollowTopic(data)),
+    closeModal: (isNew) => isNew ? dispatch(closeWireModal()) : dispatch(closeProfileModal()),
```
petrjasek: imo both do the same thing, so here the logic is not required

petrjasek:
makes perfect sense

PR #76, opened 2017-10-24
Buttons

darconny:
- buttons refactoring

### DIFF
--- a/newsroom/templates/design_companies.html
+++ b/newsroom/templates/design_companies.html
@@ -29,14 +29,14 @@
                     <span class="search__clear">
                         <img src="{{ url_for('static', filename = 'search_clear.png') }}" width="16" height="16"/>
                     </span>
-                    <button class="btn" type="submit">Search</button>
+                    <button class="aap-button" type="submit">Search</button>
                 </div>
 
             </form>
         </div>
     </div>
     <div class="content-bar__right">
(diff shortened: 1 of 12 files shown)


newsroom/templates/design_companies.html
```diff
@@ -29,7 +29,7 @@
                     <span class="search__clear">
                         <img src="{{ url_for('static', filename = 'search_clear.png') }}" width="16" height="16"/>
                     </span>
-                    <button class="btn" type="submit">Search</button>
+                    <button class="wire-button" type="submit">Search</button>
```
petrjasek: imo this class should either not be used outside wire or there should be no wire prefix, otherwise this is just confusing..

darconny: We need a class for buttons in general, that is different from .btn (bootstrap). If this is confusing lets figure out other name. Not sure about in and out wire logic, this kind of buttons are the same within whole app.

petrjasek: it would be better if you override .btn then imo, in general we can dump bootstrap, or use just parts of it. if you stay compatible with bootstrap on class names we can use their docs..

darconny:
I am closing this PR as we are going to include Bootstrap SASS components and adopt to class names as in Bootstrap.

PR #79, opened 2017-10-25
user profile base

aleksandarjelicic:
This is not finished, it's just a base which I have to commit because
Darko is making major css structure changes


aleksandarjelicic:
@akintolga @petrjasek Please don't implement this yet, it's just a base. Will let you know once I finish.

PR #80, opened 2017-10-25
updates to previous versions

petrjasek:
- hide previous versions for killed items
- if there is a newer version display it on item detail
- hide killed items from search results

### DIFF
--- a/newsroom/wire/views.py
+++ b/newsroom/wire/views.py
@@ -41,8 +41,7 @@ def get_view_data():
 
 def get_previous_versions(item):
     if item.get('ancestors'):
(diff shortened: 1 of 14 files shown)


newsroom/wire/views.py
```diff
@@ -145,6 +144,8 @@ def versions(_id):
 @blueprint.route('/wire/<_id>')
 def item(_id):
```
akintolga: If the users have the links to the killed or updated items, will we still return it?

petrjasek: right now yes. if the item is killed it will have the kill notice, but if an update is killed or previous version it will work as usual

PR #82, opened 2017-10-26
Boostrap integration, buttons refactoring

darconny:


petrjasek:
hey, not sure it's good idea to copy bootstrap files into our repo, should be possible to import just the parts you want to use, otherwise it would be kind of hard to update it when there is an update..
check http://getbootstrap.com/docs/4.0/getting-started/webpack/#importing-styles

darconny:
Hey @petrjasek , we are not sure if its good to update Bootstrap, as this could break UI at some point. For sure if it needs to be updated this should be done carefully and not automatically. I will now try to include via npm and then make cutsom.scss to override some styles, and see if this works in our case. For general discussion lets meet tomorrow and discuss use of frameworks within the application.

petrjasek:
ok sounds good

PR #86, opened 2017-10-30
add company services configuration

petrjasek:
and let users only use services from their company

### DIFF
--- a/newsroom/wire/search.py
+++ b/newsroom/wire/search.py
@@ -2,6 +2,8 @@
 import newsroom
 from flask import json
 from eve.utils import ParsedRequest
+from newsroom.auth import get_user
+from newsroom.companies import get_user_company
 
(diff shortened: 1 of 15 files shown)


newsroom/wire/search.py
```diff
@@ -46,6 +48,12 @@ def get(self, req, lookup):
             }
         }
 
+        user = get_user()
+        company = get_user_company(user)
+        if company and company.get('services'):
+            services = [code for code, is_active in company['services'].items() if is_active]
+            query['bool']['must'].append({'terms': {'service.code': services}})
```
akintolga: So if user has no company there's no filtering for now.

petrjasek: yep exactly, I keep it simple now for testing, but this won't be the case eventually

PR #91, opened 2017-11-01
fix picture renditions handling

petrjasek:
- test if it exists before usage
- add util functions to get certain renditions


petrjasek:
merging it to avoid blank page on nr-master

PR #93, opened 2017-11-01
put url params updates to utils

petrjasek:


assets/utils.js
```diff
@@ -164,3 +164,22 @@ export function toggleValue(items, value) {
     const without = items.filter((x) => value !== x);
     return without.length === items.length ? without.concat([value]) : without;
 }
+
+export function updateRouteParams(updates, state) {
```
akintolga: I think I can use this to dispact single item details view. So maybe we don't need react-router. What do you think?

petrjasek: yep I think we don't need it atm, would make more sense for settings where there is some top level routing but not in wire

PR #94, opened 2017-11-01
implement infinite scroll for wire

petrjasek:


petrjasek:
plus a bit of ui tweaks for search

PR #96, opened 2017-11-02
[SDAN-71] - Followed topic email notifications

akintolga:
- Initial `My Profile` page
- User level `Receive Emails` setting
- Emails will be sent for every matching story for the followed topic if this setting is on

### DIFF
--- a/assets/user-profile/components/UserProfileApp.jsx
+++ b/assets/user-profile/components/UserProfileApp.jsx
@@ -43,12 +43,12 @@ class UserProfileApp extends React.Component {
         let errors = {};
 
         if (!this.props.user.first_name) {
-            errors.first_name = ['Please provide first name'];
+            errors.first_name = [gettext('Please provide first name')];
             valid = false;
         }
 
         if (!this.props.user.last_name) {
-            errors.last_name = ['Please provide last name'];
+            errors.last_name = [gettext('Please provide last name')];
             valid = false;
         }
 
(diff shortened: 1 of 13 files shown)


assets/user-profile/components/UserProfileApp.jsx
```diff
@@ -32,6 +38,34 @@ class UserProfileApp extends React.Component {
         }
     }
 
+    isFormValid() {
+        let valid = true;
+        let errors = {};
+
+        if (!this.props.user.first_name) {
+            errors.first_name = ['Please provide first name'];
```
petrjasek: gettext missing

assets/user-profile/components/UserProfileApp.jsx
```diff
@@ -32,6 +38,34 @@ class UserProfileApp extends React.Component {
...
+            valid = false;
+        }
+
+        if (!this.props.user.last_name) {
+            errors.last_name = ['Please provide last name'];
```
petrjasek: same here

assets/user-profile/reducers.js
```diff
@@ -44,6 +45,16 @@ export default function itemReducer(state = initialState, action) {
             user: action.user,
         };
     }
+
+    case EDIT_USER: {
```
petrjasek: I'm not sure if field changes should make it to the store, imo it's  more like internal state of the user profile component

akintolga:
Added a new field `editedUser` for any changes that user can do

PR #99, opened 2017-11-02
allow multiple services/filters selected

petrjasek:
SDAN-110


newsroom/wire/search.py
```diff
@@ -41,6 +41,13 @@ def get_aggregation_field(key):
     return aggregations[key]['terms']['field']
 
 
+date_map = {
+    'today': 'now/d',
+    'week': 'now-1w/d',
+    'month': 'now-1M/d',
+}
```
akintolga: It looks like `date-map` is not being used.

PR #102, opened 2017-11-03
sort buckets in search facets

petrjasek:


newsroom/wire/search.py
```diff
@@ -11,9 +11,9 @@
 
 
 aggregations = {
-    'genre': {'terms': {'field': 'genre.name'}},
-    'service': {'terms': {'field': 'service.name'}},
-    'subject': {'terms': {'field': 'subject.name'}},
+    'genre': {'terms': {'field': 'genre.name', 'size': 10}},
```
akintolga: So we'll only list the top 10 genres, is that right?

petrjasek: that's the default value, we should list them all

petrjasek:
- updated numbers
- filtering out services in navigation tab without any results

PR #104, opened 2017-11-03
truncate text preview in list to 50 words

petrjasek:


aleksandarjelicic:
@petrjasek why only 50? You added it as an example for truncate functionality?
In design it has at least 3 lines of text. It should be 350 or 450.

petrjasek:
because in the spec it's 40 but that seemed too short https://dev.sourcefabric.org/browse/SDAN-41

petrjasek:
reading it again - my bad, should be 50 words not characters :wink:

aleksandarjelicic:
np, it's fine now.

PR #105, opened 2017-11-03
remove sort dropdown from searchbar

petrjasek:


aleksandarjelicic:
@petrjasek sortbar was in design. Who requested removal, aap or it's a technical issue?

petrjasek:
there is no usecase afaik and I don't want to keep it there idle

PR #108, opened 2017-11-03
Css improvements

aleksandarjelicic:


assets/styles/index.scss
```diff
@@ -851,6 +871,7 @@ article.list {
     background: #fafafa;
     box-shadow: 0px 1px 5px rgba(0, 0, 0, 0.2);
     transition: all .2s ease-out;
+    z-index: -1;
```
petrjasek: this prevents onClick events so you can't click on item previous versions which is required

aleksandarjelicic: Fixed

PR #113, opened 2017-11-04
[SDAN-124] - Display the settings icon only for admins

akintolga:


newsroom/templates/layout.html
```diff
@@ -34,7 +34,9 @@
 
         {% if session.get('name') %}
         <div class="form-inline my-lg-0 ml-auto p-2">
-            <a class="nav-link" href="{{ url_for('users.settings') }}">{{ gettext('Settings') }}</a>
+            {% if session.get('user_type') == 'administrator' %}
```
petrjasek: would probably rather introduce a template helper `is_admin` or something, so we don't duplicate that logic

PR #114, opened 2017-11-05
Notifications to use company subscriptions

akintolga:


newsroom/wire/search.py
```diff
@@ -150,8 +154,16 @@ def test_new_item(self, item_id, topics):
             }
         }
 
+        queried_topics = []
+
         for topic in topics:
-            filter = {
+            query['bool']['must'] = [{'term': {'_id': item_id}}]
```
petrjasek: must is already defined, isn't it?

newsroom/wire/search.py
```diff
@@ -163,7 +175,15 @@ def test_new_item(self, item_id, topics):
                     ],
                 }
             }
-            aggs['topics']['filters']['filters'][str(topic['_id'])] = filter
+
+            company = companies.get(str(user.get('company', '')))
+
+            # for now even if there's no active company matching for the user
+            # continuing with the search
+            set_service_query(topic_filter, company)
```
akintolga: This is per topic so line 160 affectively resets the `must` for every topic

PR #115, opened 2017-11-05
[SDAN-12] - Share a followed topic

akintolga:
- It shares the topic as a url `/?q={topic.query}` in the email

### DIFF
--- a/assets/user-profile/actions.js
+++ b/assets/user-profile/actions.js
@@ -112,7 +112,7 @@ export function fetchTopics() {
 }
 
 export function editTopic(topic) {
-    return renderModal('followTopic', topic);
+    return renderModal('followTopic', {topic});
 }
 
 /**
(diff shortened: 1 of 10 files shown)


assets/user-profile/actions.js
```diff
@@ -131,13 +131,40 @@ export function deleteTopic(topic) {
...
+ * Start share followed topic - display modal to pick users
+ *
+ * @return {function}
+ */
+export function shareTopic(items) {
+    return (dispatch, getState) => {
+        const user = getState().user;
+        const company = getState().company;
+        return server.get(`/companies/${company}/users`)
+            .then((users) => users.filter((u) => u._id !== user._id))
```
petrjasek: afaik state.user is just id, not an object

akintolga: In `user-profile` app it is an object

PR #116, opened 2017-11-05
Added missing space in previous versions

akintolga:
[SDAN-118]
[SDAN-117]
[SDAN-120]


assets/wire/components/ListItemPreviousVersions.jsx
```diff
@@ -54,7 +54,7 @@ class ListItemPreviousVersions extends React.Component {
 
         return (
             this.props.item.ancestors ?
-                <div className={`${this.baseClass}__versions`}>
+                <div className={`${this.baseClass}__versions`} id='previous-versions'>
```
petrjasek: there can be more of these visible at the same time in the list, also in the list and preview, so not sure id would work

PR #117, opened 2017-11-05
Fix for the default company option

akintolga:


assets/components/SelectInput.jsx
```diff
@@ -29,7 +29,7 @@ SelectInput.propTypes = {
     name: PropTypes.string.isRequired,
     label: PropTypes.string.isRequired,
     onChange: PropTypes.func.isRequired,
-    defaultOption: PropTypes.string,
+    defaultOption: PropTypes.bool,
```
petrjasek: this won't allow you to set default value, maybe the condition should be like `{defaultOption != null && ...}`

akintolga:
The original problem was when an empty string is provided as the defaultOption so now I have kept the bool prop as `hasDefaultOption` and added a string prop `defaultOption`.

petrjasek:
yep I know, imo using `defaultOption != null` would be enough, because that would work for `''` but not if you omit defaultOption

PR #120, opened 2017-11-06
fix filter bucket sorting for numbers

petrjasek:
urgency breaks it atm


petrjasek:
plus drop inactive service filtering from navigation bar

PR #128, opened 2017-11-07
Displaying the notification list from the top menu notification icon

akintolga:
- Styles on the pop-over are temporary until the design is ready

### DIFF
--- a/assets/components/NotificationList.jsx
+++ b/assets/components/NotificationList.jsx
@@ -2,18 +2,64 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
 
+import {gettext, shortDate} from 'utils';
+
 class NotificationList extends React.Component {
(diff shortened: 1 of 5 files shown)


petrjasek:
not sure if we should display notification about new items in the current list on top, afaik there should be refresh button with a number above current list.
also it doesn't work as notifications, because now if you don't open the notifications list while there are new items and you reload the page I guess it gets lost, while notifications should stay until you clear it.

petrjasek:
and ideally would put this notifications list into its own app, shouldn't be part of wire

akintolga:
I see your points. Probably we need more clarification about this section:
https://wiki.sourcefabric.org/display/AAPNEWS/UI+Design?preview=/65044524/77824051/AAP_Newsroom_Newswire_Notifications_2.png

akintolga:
Closing as I'm creating a new one to cover the notifications

PR #143, opened 2017-11-09
Multi action bar - new icon buttons applied

darconny:


assets/wire/components/SelectedItemsBar.jsx
```diff
@@ -16,12 +16,12 @@ class SelectedItemsBar extends React.Component {
 
     render() {
         const actions = this.props.actions.map((action) => (
-            <a className='multi-action-bar__icon' href='#'
+            <span className='icon-button'
```
petrjasek: can you make these buttons? seems like it works like button

darconny:
@petrjasek yes it makes sense, I replaced span with button tag.

PR #148, opened 2017-11-10
User Profile - Followed Topics view

aleksandarjelicic:


aleksandarjelicic:
Badge notification is not looking good, but it's a component in bootstrap that needs to change, will do that separately.

PR #149, opened 2017-11-12
Top menu notifications

akintolga:
- Notifications are sent for updates of stories that user has either downloaded or bookmarked for now
- Created a new global NotificationsApp 
- Notifications are persisted
- Initial mark up for the notifications list


newsroom/wire/search.py
```diff
@@ -215,7 +223,58 @@ def test_new_item(self, item_id, topics, users, companies):
...
 
         return topic_matches
+
+    def get_items(self, item_ids):
+        try:
+            query = {
+                'bool': {
+                    'must_not': [
+                        {'term': {'type': 'composite'}},
+                        {'constant_score': {'filter': {'exists': {'field': 'nextversion'}}}},
```
petrjasek: because of this filter searching for ancestors won't do much

webpack.config.js
```diff
@@ -7,10 +7,10 @@ const NODE_MODULES = process.env.NODE_MODULES || 'node_modules';
 module.exports = {
     entry: {
         newsroom_js: './assets/index.js',
-        settings_js: ['babel-polyfill', './assets/settings/index.js'],
-        user_profile_js: ['babel-polyfill', './assets/user-profile/index.js'],
+        settings_js: ['babel-polyfill', './assets/settings/index.js', './assets/notifications/index.js'],
+        user_profile_js: ['babel-polyfill', './assets/user-profile/index.js', './assets/notifications/index.js'],
         newsroom_css: './assets/style.js',
-        wire_js: ['babel-polyfill', './assets/wire/index.js'],
+        wire_js: ['babel-polyfill', './assets/wire/index.js', './assets/notifications/index.js'],
```
petrjasek: would rather create a new entrypoint for notifications and add it to layout, any reason to combine it?

newsroom/wire/search.py
```diff
@@ -215,7 +223,58 @@ def test_new_item(self, item_id, topics, users, companies):
...
+        search_results = self.get_items(item_ids)
+
+        if not search_results:
+            return bookmark_users
+
+        for result in search_results.hits['hits']['hits']:
+            bookmarks = result['_source'].get('bookmarks', [])
+            for bookmark in bookmarks:
+                user = users.get(bookmark)
+                if user and str(user.get('company', '')) in companies:
```
petrjasek: not sure if this companies check is needed, this would make old bookmarks stop working if user joins/changes a company, not sure if that's required?

newsroom/notifications/notifications.py
```diff
@@ -0,0 +1,58 @@
...
+            except (werkzeug.exceptions.Conflict, pymongo.errors.BulkWriteError):
+                original = super().find_one(req=None, _id=id)
+                super().update(id=id, updates={'created': now}, original=original)
+
+
+def get_user_notifications(user_id):
+    ttl = app.config.get('NOTIFICATIONS_TTL', 1)
+    lookup = {
+        'user': user_id,
+        '_created': {'$gte': utcnow() - datetime.timedelta(days=ttl)}
```
petrjasek: this should probably use `created` because that gets updated

newsroom/notifications/notifications.py
```diff
@@ -0,0 +1,58 @@
...
+
+    schema = {
+        '_id': {type: 'string', 'unique': True},
+        'item': newsroom.Resource.rel('items'),
+        'user': newsroom.Resource.rel('users'),
+        'created': {'type': 'dict', 'nullable': True},
+    }
+
+    datasource = {
+        'default_sort': [('_created', -1)]
```
petrjasek: here it should be imo `created` so it could use the index

newsroom/history.py
```diff
@@ -50,5 +49,19 @@ def transform(item):
...
 
+def get_history_users(item_ids, users, companies):
+
+    lookup = {'item': {'$in': item_ids}}
+
+    histories = query_resource('history', lookup=lookup, max_results=200)
+
+    user_matches = []
+    for history in histories:
+        if str(history.get('user', '')) in users and str(history.get('company', '')) in companies:
```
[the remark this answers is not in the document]
akintolga: Sorry the names of the variables are a bit misleading. They are active users and active companies. Here checking if the user and the company of the history item are active.

newsroom/wire/search.py
akintolga: This is to check if the company of the user is active or not. It will still work if user changes a company, as long as company is active. The name of companies should be active_companies as in this form it looks confusing.

webpack.config.js
akintolga: No good reason, let me try that.

newsroom/flaskapp.py
```diff
@@ -111,6 +111,7 @@ def _setup_jinja(self):
         self.add_template_global(self.sidenavs, 'sidenavs')
         self.add_template_global(newsroom_config)
         self.add_template_global(is_admin)
+        self.add_template_global(get_initial_notifications)
```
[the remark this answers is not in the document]
akintolga: Moved it

assets/notifications/reducers.js
```diff
@@ -15,7 +15,7 @@ export default function notificationReducer(state = initialState, action) {
     switch (action.type) {
 
     case NEW_NOTIFICATION: {
-        const notifications = state.notifications ?
+        const notifications = state.notifications.length ?
```
petrjasek: this could be also `const notifications = notifications.concat([action.notification.item]);`

newsroom/history.py
```diff
@@ -49,15 +49,15 @@ def transform(item):
...
+def get_history_users(item_ids, active_users, active_companies):
 
     lookup = {'item': {'$in': item_ids}}
 
     histories = query_resource('history', lookup=lookup, max_results=200)
 
     user_matches = []
     for history in histories:
-        if str(history.get('user', '')) in users and str(history.get('company', '')) in companies:
+        if str(history.get('user', '')) in active_users and str(history.get('company', '')) in active_companies:
```
petrjasek: wouldn't it be better to have the lookup like: `{'item': {'$in': item_ids}, 'user': {'$in': active_users}}` ?

PR #161, opened 2017-11-16
[SDAN-137] - Marking stories as read

akintolga:


assets/wire/utils.js
```diff
@@ -1,7 +1,35 @@
+/* global require */
+
+const store = require('store');
```
petrjasek: why not using import here?

assets/wire/reducers.js
```diff
@@ -173,17 +174,31 @@ export default function wireReducer(state = initialState, action) {
...
             ...state,
-            previewItem: action.item || null,
+            readItems,
+            previewItem: action.item ? action.item._id : null,
         };
+    }
+
+    case OPEN_ITEM:{
+        const readItems = Object.assign({}, state.readItems);
+        readItems[`${action.item._id}_${action.item.version}`] = true;
```
petrjasek: this would add record for each item version, what about using `{item: version}` ?

assets/wire/reducers.js
```diff
@@ -173,17 +175,34 @@ export default function wireReducer(state = initialState, action) {
             activeItem: action.item || null,
         };
 
-    case PREVIEW_ITEM:
+    case PREVIEW_ITEM: {
+        const readItems = Object.assign({}, state.readItems);
```
petrjasek: this `readItems` handling is duplicated here and in OPEN_ITEM case, could use a single function to avoid duplication

assets/wire/actions.js
```diff
@@ -20,17 +20,30 @@ export function setActive(item) {
...
 export const PREVIEW_ITEM = 'PREVIEW_ITEM';
-export function previewItem(item) {
+export function preview(item) {
     return {type: PREVIEW_ITEM, item};
 }
 
+export function previewItem(item) {
+    return (dispatch, getState) => {
+        if (item) {
+            markItemRead(item._id, getMaxVersion(getState().readItems[item._id], item.version));
```
petrjasek: imo would be better to have `markItemRead(item, getState())` and handle this version check in utils, otherwise it's duplicated. other option would be to add a function here which would be used in both previewItem and openItem

assets/wire/components/ItemsList.jsx
```diff
@@ -102,14 +103,15 @@ class ItemsList extends React.Component {
...
+        const {items, itemsById, activeItem, selectedItems, readItems} = this.props;
         const filteredActions = this.filterActions();
         const articles = items.map((_id) =>
             <WireListItem
                 key={_id}
                 item={itemsById[_id]}
                 isActive={activeItem === _id}
-                isSelected={this.props.selectedItems.indexOf(_id) !== -1}
+                isSelected={selectedItems.indexOf(_id) !== -1}
+                isRead={readItems[_id] === (parseInt(itemsById[_id].version, 10) || 0)}
```
petrjasek: would probably add another util function to keep this parseInt handling in one place

akintolga:
Thanks Petr, I refactored those duplications as well. I have also added a simple style for `--visited` stories for now.

PR #173, opened 2017-11-20
Adding short cut action buttons

akintolga:
[SDAN-135]


aleksandarjelicic:
@akintolga we need tooltips for those buttons, check here: https://nr-master.test.superdesk.org/design/wire

Also if item is bookmarked it should have icon with class icon--bookmark-remove

akintolga:
@aleksandarjelicic Thanks for the reminder, I'll check those points before merging. I'll also ping you on slack if I have any questions.

assets/components/ActionButton.jsx
```diff
@@ -0,0 +1,29 @@
...
+import PropTypes from 'prop-types';
+
+
+function ActionButton({item, className, displayName, action}) {
+    return (
+        <button
+            type='button'
+            className={className}
+            onClick={() => action.action(action.multi ? [item._id] : item)}
+            ref={(elem) => $(elem).tooltip()}
```
petrjasek: this will also need some [dispose](http://getbootstrap.com/docs/4.0/components/tooltips/#tooltipdispose) call otherwise it would leak memory. would be probably better also to call `.tooltip()` on `componentDidMount` event

PR #183, opened 2017-11-25
[SDAN-153] - Making visited buttons grayed

akintolga:
Depends on https://github.com/superdesk/superdesk-core/pull/1097

### DIFF
--- a/assets/wire/reducers.js
+++ b/assets/wire/reducers.js
@@ -102,7 +102,7 @@ function updateItemActions(state, items, action) {
 
     items.map((item) => {
         itemsById[item] = Object.assign({}, itemsById[item]);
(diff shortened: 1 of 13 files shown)


assets/wire/reducers.js
```diff
@@ -93,6 +97,17 @@ function getReadItems(state, item) {
     return readItems;
 }
 
+function updateItemActions(state, items, action) {
+    const itemsById = Object.assign({}, state.itemsById);
+
+    items.map((item) => {
+        itemsById[item] = Object.assign({}, itemsById[item]);
+        itemsById[item][action] = (itemsById[item].bookmarks || []).concat([state.user]);
```
petrjasek: not sure why it's using bookmarks here?

akintolga: That doesn't look good, fixing now...

PR #198, opened 2017-12-03
Newsroom products

akintolga:
- [ ] Create/Edit/Remove products
- [ ] Assign products to companies
- [ ] Filter content by products

### DIFF
--- a/assets/settings/components/SettingsApp.jsx
+++ b/assets/settings/components/SettingsApp.jsx
@@ -9,8 +9,10 @@ import {
     postItem,
     deleteItem,
     newItem,
+    newProduct,
     updateMenu,
     fetchItems,
+    fetchProducts,
     fetchCompanyUsers,
 } from '../actions';
 import UserBar from './users/UserBar';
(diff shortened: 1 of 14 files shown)


assets/settings/components/SettingsApp.jsx
```diff
@@ -86,6 +91,32 @@ class SettingsApp extends React.Component {
...
 
+                        />
+                    </div> : null }
+                {this.props.selectedMenu === 'products' ?
+                    <div className="content">
+                        <ProductBar
+                            onNewProduct={this.props.newProduct}
+                            fetchItems={this.props.fetchItems}
+                        />
+                        <Products
```
petrjasek: imo settings app should handle the routing/tabs, but should not be aware about all those Products properties. pls use connect with Products and handle it there

assets/settings/reducers.js
```diff
@@ -26,6 +28,7 @@ const initialState = {
     activeQuery: null,
     selectedMenu: 'companies',
     services: [],
+    productsById: {},
```
petrjasek: would be good imo if we put this into `products` namespace so we can combine reducers by settings app instead of having it all together

newsroom/products/products.py
```diff
@@ -0,0 +1,60 @@
+
+import newsroom
+import superdesk
+
+from content_api import MONGO_PREFIX
```
petrjasek: `newsroom.Resouce` handles the prefix

petrjasek:
what's the idea regarding products hierarchy? imo we don't need it, but we should introduce another entity for the top level navigation (categories or something) and make products (wire in this case) related to it. so we can have different filtering products for agenda/wire/events but they could be all related to some top level category

PR #204, opened 2017-12-05
Newsroom products

akintolga:
- [x] Separate apps for settings menu items
- [x] Create/Edit/Remove products and navigations
- [x] Assign products to companies and to navigations
- [x] Filter content by products


akintolga:
Few notes on the changes:
- Navigation item will appear for users if there's at least one active product associated with it. (I was originally planning to use aggregations)
- `Services` are now replaced by `Navigations`
- If there's no products associated for a company then users won't see any content

akintolga:
Related to https://github.com/superdesk/superdesk-core/pull/1108

assets/navigations/components/NavigationBar.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { gettext } from 'utils';
+import SearchBar from 'wire/components/SearchBar';
+
+class NavigationBar extends React.Component {
```
petrjasek: imo there should be a component for all those SomethingBars, atm it's duplicated everywhere with different label and callbacks

assets/navigations/components/Navigations.jsx
```diff
@@ -0,0 +1,113 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import EditNavigation from './EditNavigation';
+import NavigationList from './NavigationList';
+import SearchResultsInfo from 'wire/components/SearchResultsInfo';
+import {setError} from '../actions';
+import {gettext} from 'utils';
+
+class Navigations extends React.Component {
```
petrjasek: given that this is connected component would be better for encapsulation if you use map state/dispatch to props here rather than on the NavigationsApp which is only passing it all down without having any state

newsroom/templates/settings.html
```diff
@@ -7,14 +7,57 @@
...
 {% endblock %}
 
 {% block script %}
 <script>
 var viewData = {{ data | tojson | safe }};
 </script>
 
-{{ javascript_tag('settings_js') | safe }}
+    {% if setting_type == "companies" %}
+        {{ javascript_tag('companies_js') | safe }}
```
petrjasek: imo it could be `{{ javascript_tag('%s_js' % setting_type) | safe }}`

newsroom/products/products.py
```diff
@@ -0,0 +1,67 @@
...
+        'query': {
+            'type': 'string'
+        },
+        'is_enabled': {
+            'type': 'boolean',
+            'default': True
+        },
+        'product_type': {
+            'type': 'string',
+            'allowed': ['query', 'superdesk', 'curated']
```
petrjasek: why types? wouldn't it be easier to config if you could combine superdesk id with a query?

newsroom/navigations/navigations.py
```diff
@@ -0,0 +1,75 @@
+import newsroom
+import superdesk
+
+
+class NavigationsResource(newsroom.Resource):
+    """
+    Navigations schema
+    """
+    schema = {
```
petrjasek: imo it's missing some field we could use for manual sorting

assets/wire/reducers.js
```diff
@@ -111,16 +112,21 @@ function updateItemActions(state, items, action) {
 
 function _wireReducer(state, action) {
     switch (action.type) {
-    case TOGGLE_SERVICE: {
-        const activeService = {};
-        if (action.service) {
-            activeService[action.service.code] = true;
+
+    case TOGGLE_NAVIGATION: {
```
petrjasek: imo it would be better if we only deal with navigation on client side and resolve products on server. it's an implementation detail we don't have to expose

assets/wire/reducers.js
```diff
@@ -61,8 +61,9 @@ const initialState = {
     newItemsByTopic: {},
     readItems: {},
     wire: {
-        services: [],
-        activeService: {},
+        navigations: [],
+        activeProducts: {},
+        activeNavigation: {},
```
petrjasek: I think there will be only 1 active nav at a time so we can simplify the state, we were testing multiple products active but it didn't work out

assets/wire/reducers.js
akintolga: Done that

newsroom/navigations/navigations.py
akintolga: Added `order` field

assets/companies/components/CompaniesApp.jsx
```diff
@@ -75,30 +52,11 @@ CompaniesApp.propTypes = {
...
-    isLoading: state.isLoading,
-    activeQuery: state.activeQuery,
-    totalCompanies: state.totalCompanies,
-    companyOptions: state.companyOptions,
-    companiesById: state.companiesById,
-    companyUsers: state.companyUsers,
-    errors: state.errors,
-});
 
 const mapDispatchToProps = (dispatch) => ({
```
petrjasek: fyi this could be a dict and then it wraps all it's values into dispatch, so this could be:
`const mapDispatchToProps = {
    newCompany,
    fetchCompanies,
    setQuery,
};
```

File: newsroom/navigations/navigations.py [hunk 9]
```diff
@@ -18,11 +18,15 @@ class NavigationsResource(newsroom.Resource):
...
+        },
+        'order': {
+            'type': 'integer',
+            'nullable': 'true'
         }
     }
     datasource = {
         'source': 'navigations',
-        'default_sort': [('name', 1)]
+        'default_sort': [('order', 1)]
```
petrjasek: what about `[('order', 1), ('name', 1)]`

assets/wire/reducers.js
```diff
@@ -114,19 +113,16 @@ function _wireReducer(state, action) {
     switch (action.type) {
 
     case TOGGLE_NAVIGATION: {
-        const activeProducts = {};
         const activeNavigation = {};
 
-        if (action.navigation && action.navigation.products) {
-            action.navigation.products.map((product) => activeProducts[product] = true);
+        if (action.navigation) {
             activeNavigation[action.navigation._id] = true;
```
petrjasek: would still make this `activeNavigation = action.navigation && action.navigation._id;` there is no need for this to be a dict

newsroom/navigations/views.py
```diff
@@ -0,0 +1,85 @@
...
+    return jsonify({'success': True}), 200
+
+
+@blueprint.route('/navigations/<id>', methods=['DELETE'])
+@admin_only
+def delete(id):
+    """ Deletes the navigations by given id """
+    get_entity_or_404(id, 'navigations')
+
+    # remove all references from products
```
petrjasek: imo we should probably not allow removal if there are products using it

akintolga: Maybe they can get away by disabling only. But I think they may need to delete these navigations as imo there'll few of those that come and go but not sure. I'll keep this for now.

PR #207, opened 2017-12-08
Read/unread item UI

darconny:


petrjasek:
can you avoid this commenting things out in the template? if it's not used just remove it

darconny:
@petrjasek I put this code back in, as we need it. It was temporarily hidden.

petrjasek:
I see

PR #209, opened 2017-12-12
Fix for products with empty companies

akintolga:


newsroom/products/products.py
```diff
@@ -54,10 +54,10 @@ def get_products_by_navigation(navigation_id):
 def get_products_by_company(company_id):
     # TODO(tolga): this lookup by company id is not working
     products = list(superdesk.get_resource_service('products').get(req=None, lookup={'is_enabled': True}))
-    return [p for p in products if str(company_id) in p.get('companies')]
+    return [p for p in products if str(company_id) in p.get('companies', [])]
```
petrjasek: I think this lookup not working might be because `companies` is list of strings

akintolga: That maybe the case, I'll re-visit this area for sure.

PR #210, opened 2017-12-12
Responsive UI

darconny:
- Settings page: Users, companies


assets/components/ListBar.jsx
```diff
@@ -11,7 +11,7 @@ class ListBar extends React.Component {
                     <SearchBar setQuery={this.props.setQuery} fetchItems={()=>this.props.fetch()}/>
                     <div className="content-bar__right">
                         <button
-                            className="btn btn-outline-secondary"
+                            className="btn btn-outline-secondary btn responsive"
```
petrjasek: it's missing `-` between `btn` and `responsive`

darconny: Tnx, fixed.

PR #215, opened 2017-12-15
populate item word_count on push

petrjasek:
so there is no need to compute it on client


akintolga:
Imo having content-api in its own repo (with its own version) may be more sustainable on the long run.

PR #218, opened 2017-12-17
[SDAN-68] - Testing query in product

akintolga:


newsroom/wire/search.py
```diff
@@ -83,6 +102,9 @@ def set_product_query(query, company):
         for product in products:
             if product.get('query'):
                 query['bool']['must'][0]['bool']['should'].append(_query_string(product['query']))
+    else:
+        # user does not belong to a company so blocking all stories
+        query['bool']['must'].append({'bool': {'must': [{'terms': {'products.id': [-1]}}]}})
```
petrjasek: I think doing `abort(403)` might be better - we can handle it on client and display some message.

PR #235, opened 2018-01-17
Wire grid view

darconny:


assets/index.js
```diff
@@ -7,19 +7,41 @@ import 'whatwg-fetch';
...
+    }
+}
+
+function responsiveWireItem() {    
+
     if (filterOpen && !previewOpen) {
         document.getElementsByClassName('wire-column__main')[0].classList.add('wire-articles__one-side-pane');
         document.getElementsByClassName('wire-column__main')[0].classList.remove('wire-articles__two-side-panes');
+        if ( isGrid() ) {        
+            $('.wire-articles__item-wrap').removeClass().addClass('wire-articles__item-wrap col-sm-12 col-md-6 col-xl-4 col-xxl-3');
```
petrjasek: is this something we should be doing in react?

darconny: Yes I suppose so.

PR #241, opened 2018-01-25
Side navigation tooltips

darconny:


assets/index.js
```diff
@@ -10,6 +10,16 @@ $('.content-main').scroll(function() {
     $('.carousel-caption').css('opacity', 1 - scrollTop / 400);
 });
 
+function is_touch_device() {
```
petrjasek: this won't change much so imo would make sense to make this a constant:
```
const IS_TOUCH_DEVICE = ...;
```
but not a big deal

PR #248, opened 2018-02-02
add watermark to pictures on push binary

petrjasek:
SDAN-138


newsroom/push.py
```diff
@@ -181,3 +201,29 @@ def push_binary_get(media_id):
...
+    if image.mode != 'RGBA':
+        image = image.convert('RGBA')
+
+    with app.open_resource('static/watermark.png') as watermark_binary:
+        watermark_image = Image.open(watermark_binary)
+        set_opacity(watermark_image, 0.3)
+        watermark_layer = Image.new('RGBA', image.size)
+        watermark_layer.paste(watermark_image, (
+            image.size[0] - watermark_image.size[0],
+            int((image.size[1] - watermark_image.size[1]) * 0.66),
```
akintolga: I think watermark may be off centered with constant `0.66` but I don't think it matters.

PR #251, opened 2018-02-06
Added new card types

akintolga:


assets/home/components/LargeTextOnlyCard.jsx
```diff
@@ -0,0 +1,55 @@
...
+                <div className='wire-articles__item__meta'>
+                    <div className='wire-articles__item__meta-info'>
+                        <span className='bold'>{item.slugline}</span>
+                        <span>{gettext('Source: {{ source }}', {source: item.source})}
+                            {' // '} {shortDate(item.versioncreated)}
+                        </span>
+                    </div>
+                </div>
+            </div>
+            <div className='card-footer'>
```
petrjasek: imo would be good to introduce some components for card parts, like CardFooter or CardBody seems like it's duplicated there

PR #253, opened 2018-02-06
fix picture usage in home widgets

petrjasek:
there will be thumbnail and viewImage rendition without watermark,
those should be used for wire list and home views with small images.

so utils.getThumbnailRendition will be without watermark, utils.getPreviewRendition
should have it.

SDAN-138


akintolga:
I probably need to do the same to the existing PR for two new cards

PR #262, opened 2018-02-18
[SDAN-179] - Modifications for picture-text card

akintolga:
Also added abstract to 6 x text-only cards [SDAN-181]


akintolga:
@darconny Please take a look at the changes when you have a chance (preview is [here](http://nrpr-262.test.superdesk.org))

PR #266, opened 2018-02-24
[SDAN-194] - Home page carousel

akintolga:


newsroom/templates/home.html
```diff
@@ -16,4 +16,15 @@
         homeData = {{ data | tojson | safe }};
     </script>
     {{ javascript_tag('home_js') | safe }}
-{% endblock %}
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery.touchswipe/1.6.18/jquery.touchSwipe.js" integrity="sha256-KKv6+21Uvs0xGccvranH3/RaUvz7N24khCbpW6leF5o=" crossorigin="anonymous"></script>
+    <script>
+        // Carousel swipe
+        $('.carousel').swipe({
```
petrjasek: would it be possible to handle this in the react component?

akintolga: @petrjasek I'm afraid I used wrong component to wire the app photos. I just came to know that toip carousel is been ditched, I nede to use the 4 x gallery cards instead. So `swipe` will go away. I'll push another commit for this soon.

PR #269, opened 2018-02-27
Using the square image link for the photo card

akintolga:


petrjasek:
it looks a bit stretched there, there is some issue with css minHeight

akintolga:
Yes, I modified min-height to prevent streching

PR #272, opened 2018-02-27
Event management

akintolga:
SDAN-190

### DIFF
--- a/assets/cards/components/Cards.jsx
+++ b/assets/cards/components/Cards.jsx
@@ -31,7 +31,13 @@ class Cards extends React.Component {
         let errors = {};
 
         if (!this.props.cardToEdit.label) {
-            errors.name = ['Please provide card label'];
+            errors.label = ['Please provide card label'];
+            valid = false;
+        }
+
+        if (this.props.cardToEdit.type === '2x2-events' &&
+          !this.props.cardToEdit.config.events[0].startDate) {
+            errors.event_0_startDate = ['Please provide start date'];
             valid = false;
         }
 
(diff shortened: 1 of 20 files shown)


assets/cards/components/Cards.jsx
```diff
@@ -31,7 +31,13 @@ class Cards extends React.Component {
         let errors = {};
 
         if (!this.props.cardToEdit.label) {
-            errors.name = ['Please provide card label'];
+            errors.label = ['Please provide card label'];
```
petrjasek: gettext missing

assets/cards/components/Cards.jsx
```diff
@@ -31,7 +31,13 @@ class Cards extends React.Component {
...
+            valid = false;
+        }
+
+        if (this.props.cardToEdit.type === '2x2-events' &&
+          !this.props.cardToEdit.config.events[0].startDate) {
+            errors.event_0_startDate = ['Please provide start date'];
```
petrjasek: gettext missing

newsroom/utils.py
```diff
@@ -41,6 +45,14 @@ def get_entity_or_404(_id, resource):
     return item
 
 
+def get_file(key):
+    file = request.files.get(key)
+    if file:
+        filename = secure_filename(file.filename)
+        file.save(os.path.join(app.config['UPLOAD_FOLDER'], filename))
```
petrjasek: we should store the files to `app.media`, otherwise we would lost them on each deployment

PR #274, opened 2018-02-27
use gunicorn as app server

petrjasek:


Vagrantfile
```diff
@@ -1,49 +1,41 @@
...
+honcho start -p 5050 &
 SCRIPT
 
 Vagrant.configure("2") do |config|
   config.vm.box = "ubuntu/xenial64"
 
-  config.vm.network "forwarded_port", guest: 5000, host: 5050
-  config.vm.network "forwarded_port", guest: 5100, host: 5100
+  config.vm.network "forwarded_port", guest: 5050, host: 5050
+  config.vm.network "forwarded_port", guest: 5150, host: 5100
```
akintolga: Just double checking if these ports are ok

petrjasek: yep that's due to https://github.com/superdesk/newsroom/pull/274/files#diff-23b6f443c01ea2efcb4f36eedfea9089R31

PR #278, opened 2018-03-04
Company reports

akintolga:
A proof of concept for company reports.

### DIFF
--- a/assets/company-reports/actions.js
+++ b/assets/company-reports/actions.js
@@ -2,6 +2,13 @@ import {errorHandler} from '../utils';
 import server from '../server';
 
 
+const REPORTS = {
+    'company-saved-searches': '/reports/company/saved_searches',
+    'user-saved-searches': '/reports/user/saved_searches',
+    'company-products': '/reports/company/products',
+};
+
+
 export const QUERY_REPORT = 'QUERY_REPORT';
 export function queryReport() {
     return {type: QUERY_REPORT};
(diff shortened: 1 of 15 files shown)


assets/company-reports/actions.js
```diff
@@ -0,0 +1,82 @@
...
+
+export const SET_ERROR = 'SET_ERROR';
+export function setError(errors) {
+    return {type: SET_ERROR, errors};
+}
+
+export function runReport() {
+    return function (dispatch, getState) {
+        dispatch(queryReport());
+        if (getState().activeReport == 'company-saved-searches') {
```
petrjasek: this seems like mapping activeReport -> url so we could have a const dict:
```
const REPORTS = {'company-saved-searches': '/reports/componay/saved_searches', ...};
```
and later it would just get url and query:
```
const url = REPORTS[getState().activeReport];
dispatch(queryReport());
return server.get(url).then((data) => { dispatch(receivedData(data)));
```

assets/company-reports/actions.js
```diff
@@ -0,0 +1,82 @@
...
+import server from '../server';
+
+
+export const QUERY_REPORT = 'QUERY_REPORT';
+export function queryReport() {
+    return {type: QUERY_REPORT};
+}
+
+export const SET_ACTIVE_REPORT = 'SET_ACTIVE_REPORT';
+export function setActiveReport(data) {
```
petrjasek: imo would be better to pass just report name and not the event object

newsroom/templates/layout.html
```diff
@@ -56,6 +56,11 @@
                       <i class="icon--cog icon--mid-blue"></i>
                   </a>
                 </li>
+                <li class="sidenav-icons__item sidenav-icons__item--report {% if request.endpoint == 'companies.settings' %}active{% endif %}">
+                  <a href="{{ url_for('reports.company_reports') }}" title="{{ gettext('Settings') }}" data-toggle="tooltip" data-placement="right">
```
petrjasek: title should be Reports probably

assets/company-reports/actions.js
akintolga: Yes, I have done that on `onChange` function

akintolga:
I'm keeping this PR simple, I'll do another one for printing next. The icon is still outstanding, I'll check that with darko later.

PR #279, opened 2018-03-05
UI adjustments

darconny:


newsroom/templates/layout.html
```diff
@@ -24,7 +24,7 @@
 
     <nav class="main navbar justify-content-start align-items-stretch">
         <div class="navbar__logo">
-            <img src="{{ url_for('static', filename = 'logo.svg') }}" width="34" height="34"/>
+            <img src="{{ url_for('static', filename = 'logo-blue.svg') }}" width="34" height="34"/>
```
petrjasek: can you keep it `logo.svg` ? it's overriden on some branches for other users

darconny: I tried but it failed to overwrite, that's why I put separate image. Let me try again.

PR #281, opened 2018-03-07
[SDAN-197] - Hide count on the main pages

akintolga:
- the number of items should not be displayed on the "main" pages - i.e. the Wire home page, pages from navigation
- the number of items should be displayed for:
-- search results
-- when filters are applied
-- topics
-- bookmarks


akintolga:
Reopenning to kick the builds

PR #307, opened 2018-03-30
[SDAN-216] - My topics tab

akintolga:
Seperates saved topics to `My Topics` tab
Second commit in this PR enables user to open topic management pop up from wire app

### DIFF
--- a/assets/index.js
+++ b/assets/index.js
@@ -22,8 +22,7 @@ if ( !isTouchDevice() ) {
 var filterOpen = false;
 var previewOpen = false;
 
-window.manageTopics = document.createEvent('Event');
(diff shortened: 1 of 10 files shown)


assets/index.js
```diff
@@ -22,6 +22,9 @@ if ( !isTouchDevice() ) {
 var filterOpen = false;
 var previewOpen = false;
 
+window.manageTopics = document.createEvent('Event');
+window.manageTopics.initEvent('manage_topics', true, true);
```
petrjasek: seems like it's [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Event/initEvent)

PR #308, opened 2018-03-31
[SDAN-215] - Moving filters from post-filters

akintolga:
Created a new setting `FILTER_BY_POST_FILTER` to control the aggregations on the results. For AAP we can set it to `False` so they get renewed aggs for evey new filter

### DIFF
--- a/newsroom/wire/search.py
+++ b/newsroom/wire/search.py
@@ -198,7 +198,7 @@ class WireSearchService(newsroom.Service):
             filters = json.loads(req.args['filter'])
 
         if not app.config.get('FILTER_BY_POST_FILTER', False):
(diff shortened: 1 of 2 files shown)


newsroom/wire/search.py
```diff
@@ -192,11 +192,29 @@ def get(self, req, lookup):
         if req.args.get('bookmarks'):
             _set_bookmarks_query(query, req.args['bookmarks'])
 
+        filters = None
+
+        if req.args.get('filter'):
+            filters = json.loads(req.args['filter'])
+
+        if not app.config.get('FILTER_BY_POST_FILTER', False):
+            if filters and not app.config.get('FILTER_BY_POST_FILTER', False):
```
petrjasek: this should be only `if filters:` the other part is always true

PR #313, opened 2018-04-05
fix deployment related issues

petrjasek:
- mainly packaging and dependencies


petrjasek:
changes to make https://github.com/superdesk/newsroom-aap work

PR #315, opened 2018-04-05
[SDAN-218] - Add current date to the header

darconny:
@akintolga , @petrjasek I added div blocks to the header, we need js code to populate it with current time in format "Tuesday, 4/4/2018".


petrjasek:
you can put there `{{ None | dateformat('medium') }}` for now so it's not empty

PR #318, opened 2018-04-08
UI changes

akintolga:
[SDAN-221]

### DIFF
--- a/assets/config.js
+++ b/assets/config.js
@@ -1,0 +1,3 @@
+export default {
+    displaySourceOnHome: false,
+};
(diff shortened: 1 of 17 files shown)


assets/config.js
```diff
@@ -0,0 +1,3 @@
+export default {
+    displaySourceOnHome: false,
```
petrjasek: is it temporary thing?

akintolga: I was thinking that tomorrow another newsroom client may ask `source` to be `visible` so I created a simple config but if you think it is premature then I can get rid of it

petrjasek: I'm just wondering how we would change it if it's imported from a file, maybe if you would read it from `window.newsroomConfig` or something.

akintolga: Yes, that should do. After merging this PR I'll create a basic wiring for `window.newsroomConfig` to pass the configuration. Maybe we can keep this as default or something in that case.

PR #319, opened 2018-04-10
[SDAN-223] - Fixing copy function when the panel is opened

akintolga:
At the moment copy relies on dom element for preview panel to exist. This solution opens the preview panel before copy takes place.

### DIFF
--- a/assets/wire/actions.js
+++ b/assets/wire/actions.js
@@ -26,6 +26,14 @@ export function preview(item) {
     return {type: PREVIEW_ITEM, item};
 }
 
+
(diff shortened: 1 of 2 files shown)


assets/wire/actions.js
```diff
@@ -26,6 +26,17 @@ export function preview(item) {
     return {type: PREVIEW_ITEM, item};
 }
 
+export function previewAsync(item) {
+    return (dispatch) => new Promise(resolve => {
```
petrjasek: why promise?

PR #328, opened 2018-04-24
[SDAN-229] - Adding copyright page

akintolga:
- I created a new page for `copyright` instead of a pop up

### DIFF
--- a/newsroom/templates/layout.html
+++ b/newsroom/templates/layout.html
@@ -81,10 +81,10 @@
         {% endblock %}
     </div>
     <footer class="footer">
-        <a href="https://www.aap.com.au/legal/" target="_blank">Privacy Policy</a>
-        <a href="https://www.aap.com.au/legal/" target="_blank">Terms and Conditions</a>
-        <a href="https://www.aap.com.au/contact/sales-inquiries/" target="_blank">Contact Us</a> 
-        <a href="#">Copyright</a>
+        <a href="https://www.aap.com.au/legal/" target="_blank">{{ gettext('Privacy Policy') }}</a>
+        <a href="https://www.aap.com.au/legal/" target="_blank">{{ gettext('Terms and Conditions') }}</a>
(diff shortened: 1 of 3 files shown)


newsroom/wire/views.py
```diff
@@ -142,6 +142,11 @@ def search():
     return send_response('wire_search', response)
 
 
+@blueprint.route('/copyright')
```
petrjasek: I think it would make sense to have another blueprint for this, it's not clear why it's part of wire

newsroom/templates/layout.html
```diff
@@ -81,10 +81,10 @@
         {% endblock %}
     </div>
     <footer class="footer">
-        <a href="https://www.aap.com.au/legal/" target="_blank">Privacy Policy</a>
-        <a href="https://www.aap.com.au/legal/" target="_blank">Terms and Conditions</a>
-        <a href="https://www.aap.com.au/contact/sales-inquiries/" target="_blank">Contact Us</a> 
-        <a href="#">Copyright</a>
+        <a href="https://www.aap.com.au/legal/" target="_blank">{{ gettext('Privacy Policy') }}</a>
```
petrjasek: pls put put this into a `footer.html` template so we can customize it via themes easily

newsroom/templates/footer.html
```diff
@@ -0,0 +1,6 @@
+<footer class="footer">
+    <a href="https://www.aap.com.au/legal/" target="_blank">{{ gettext('Privacy Policy') }}</a>
+    <a href="https://www.aap.com.au/legal/" target="_blank">{{ gettext('Terms and Conditions') }}</a>
+    <a href="https://www.aap.com.au/contact/sales-inquiries/" target="_blank">{{ gettext('Contact Us') }}</a>
+    <a href="/copyright">{{ gettext('Copyright') }}</a>
```
petrjasek: would be good to use `{{ url_for('public.copyright') }}` so we know where it's defined

PR #329, opened 2018-04-26
Dynamic footer for the home page

akintolga:
Just a poc for displaying footer dynamically on the home page (only). The footer needs to be rendered dynamically to make it appear when scroll position at (or close) to the bottom.

### DIFF
--- a/newsroom/templates/layout.html
+++ b/newsroom/templates/layout.html
@@ -81,7 +81,11 @@
         {% endblock %}
     </div>
 
-     {% include "footer.html" %}
+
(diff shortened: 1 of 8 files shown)


newsroom/templates/layout.html
```diff
@@ -81,7 +81,11 @@
         {% endblock %}
     </div>
 
-     {% include "footer.html" %}
+
+    <footer id="footer">
+        {% block footer %}
```
petrjasek: you can add that include here and it will be used as default if not set

PR #332, opened 2018-04-29
Forgotten password page improvements

akintolga:
[SDAN-227] - Rename password reset
[SDAN-228] - Forgotten Password validation
[SDAN-230] - Forgotten password message

### DIFF
--- a/newsroom/templates/login.html
+++ b/newsroom/templates/login.html
@@ -42,10 +42,6 @@
             </div>
         </div>
     </div>
-    <div class="row">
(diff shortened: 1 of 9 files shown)


newsroom/templates/login.html
```diff
@@ -42,5 +42,10 @@ <h5 class="mb-0">{{ gettext("Login") }}</h5>
             </div>
         </div>
     </div>
+    <div class="row">
```
petrjasek: looks like this should be part of some layout page

akintolga: That requires two layouts, do you want me to create as there are two inner divs?

PR #340, opened 2018-05-07
A login endpoint to retrieve a token

akintolga:
This is an experimental change to have:
- an end point to get a token (`POST /login/token/`)
- another to login with the token. (`GET /login/token/`)

The `POST` endpoint can be used to get the valid token. This endpoint could be limited to a certain post  parameters or an ip address
The `GET` endpoint will authenticate the user and log her in


akintolga:
@mdhaman  Please let me if this would help you to login remotely

mdhaman:
@akintolga if you enable CORS. We should be able to test it.

mdhaman:
👍 look good. Had to modify `nginx` configuration.

PR #345, opened 2018-05-19
[SDAN-264] - Changes to email templates

akintolga:


petrjasek:
would be good to introduce a layout for emails so we have the footer consistent everywhere

petrjasek:
would rather use a layout so we can add a common header as well if needed, like in https://github.com/superdesk/superdesk-core/blob/master/superdesk/templates/email_layout.html

newsroom/templates/email_layout.html
```diff
@@ -0,0 +1,5 @@
+<p>{{ name }},</p>
+
+{% block content %}{% endblock %}
+
+{% include "email_footer.html" %}
```
petrjasek: now you can add the footer here instead of including it, or is it included anywhere else?

akintolga: 🥇 <-- This is what I need I guess

PR #354, opened 2018-05-29
[SDAN-269] - Broken link for the story in notification email

akintolga:


petrjasek:
what about [flask.url_for](http://flask.pocoo.org/docs/1.0/api/#flask.url_for) that should avoid such issues in future

PR #355, opened 2018-05-30
[SDAN-270] - Missing UI changes for wire page

akintolga:


akintolga:
reopenning for build to run again

PR #369, opened 2018-06-17
Ingesting events and planning items by merged `agenda`

akintolga:
- Trying to merge the event and planning items into an `agenda` entity
- Agenda will have a start and end date
- Coverages will be a `nested` property
- Original event and planning items will also be stored in the agenda object


newsroom/push.py
```diff
@@ -87,17 +121,136 @@ def publish_item(doc):
...
-    item = flask.json.loads(flask.request.get_data())
-    assert 'guid' in item, {'guid': 1}
-    assert 'type' in item, {'type': 1}
-    orig = app.data.find_one('wire_search', req=None, _id=item['guid'])
-    item['_id'] = publish_item(item)
-    notify_new_item(item, check_topics=orig is None)
-    app.cache.delete(HOME_ITEMS_CACHE_KEY)
-    return flask.jsonify({})
+def publish_event(event):
+    orig = app.data.find_one('agenda', req=None, _id=event['guid'])
```
petrjasek: I guess it should be searching by `event_id` not `_id`

newsroom/push.py
```diff
@@ -87,17 +121,136 @@ def publish_item(doc):
...
+
+def publish_planning(planning):
+    service = superdesk.get_resource_service('agenda')
+
+    # update dates
+    planning['planning_date'] = datetime.strptime(planning['planning_date'], '%Y-%m-%dT%H:%M:%S+0000')
+
+    if planning.get('event_item'):
+        # this is a planning for an event item
+        agenda = app.data.find_one('agenda', req=None, _id=planning['event_item'])
```
petrjasek: same here probably, search by `event_id` would make more sense imo

newsroom/push.py
```diff
@@ -87,17 +121,136 @@ def publish_item(doc):
...
+            return service.patch(agenda['_id'], agenda)
+
+        for date in agenda['dates']:
+            if date['start'].date() == planning['planning_date'].date():
+                date['planning_id'] = planning['guid']
+                date['coverages'] = [c['planning']['g2_content_type'] for c in planning.get('coverages', [])]
+
+    else:
+        # there's no event item (ad-hoc planning item)
+        agenda = {}
```
petrjasek: here it might search by `planning_id` - there could be updates to ad-hoc planning item

newsroom/default_settings.py
```diff
@@ -97,6 +100,8 @@
 WEBPACK_ASSETS_URL = os.environ.get('ASSETS_URL', '/static/dist/')
 WEBPACK_SERVER_URL = os.environ.get('WEBPACK_SERVER_URL', 'http://localhost:8080/')
 
+SERVER_DOMAIN = 'localhost'
```
petrjasek: why is this needed?

newsroom/push.py
akintolga: Ok, in that case this _id could be an autogenerated id for agenda I guess

newsroom/agenda/agenda.py
```diff
@@ -0,0 +1,102 @@
...
+            'type': 'dict',
+            'schema': {
+                'start': {'type': 'datetime'},
+                'end': {'type': 'datetime'},
+                'state': WORKFLOW_STATE_SCHEMA,
+                'planning_id': planning_schema['guid'],
+                'coverages': {'type': 'list', 'nullable': True},
+            }
+        },
+        'mapping': not_analyzed,
```
petrjasek: I think for the mapping we will need a [nested objects] in order to filter by start/end

newsroom/default_settings.py
akintolga: Superdesk-core was complaining about it when I used the function `get_recurring-events` so I put this in to pass that

newsroom/push.py
petrjasek: actually using _id will be probably faster, so I would keep it but you can add a comment

newsroom/push.py
akintolga: Like the one above, I'm planning to add a comment here like `if there's an event then _id field will have the same value as event_id` ...

newsroom/push.py
mdhaman: Also, you can retrospectively create an event from the ad-hoc planning item.

petrjasek: that's a good scenario, not sure what should happen then, I guess there will be another item created for the event and this item should be added to it right @akintolga ?

petrjasek: I guess it would update the planning item adding some event_id, so we should get it. but there will be also that event published eventually and I don't think it will have id for the planning item

petrjasek:
pls add it to feat-agenda branch

akintolga:
Closing this as need to reopen against feat-agenda branch

PR #371, opened 2018-06-22
Ingesting events and planning items by merged agenda

akintolga:
- Trying to merge the event and planning items into an `agenda` entity
- Agenda will have a start and end date
- Coverages will be a `nested` property
- Original event and planning items will also be stored in the agenda object

Moving the [original PR](https://github.com/superdesk/newsroom/pull/369) to feat-agenda branch


newsroom/agenda/agenda.py
```diff
@@ -0,0 +1,113 @@
...
+    schema['event_id'] = events_schema['guid']
+    schema['recurrence_id'] = {
+        'type': 'string',
+        'mapping': not_analyzed,
+        'nullable': True,
+    }
+
+    # content metadata
+    schema['name'] = events_schema['name']
+    schema['slugline'] = {'type': 'string'}
```
petrjasek: I think we should use `not_analyzed` here

newsroom/agenda/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+from newsroom.agenda.agenda import AgendaResource, AgendaService
+from superdesk.metadata.item import not_analyzed
+
+blueprint = Blueprint('agenda', __name__)
+
+from . import views  # noqa
+
+
+def init_app(app):
+    app.config['DOMAIN']['events']['schema'].update({
```
petrjasek: this would be better done in planning

PR #384, opened 2018-07-05
SDAN-262 / SDAN-281

darconny:
Fix for both issues, as with merging 281 fix search bar was hidden with
scrolling the list.


silvalooky:
can somone please help me to install newsroom ?

PR #407, opened 2018-07-17
fix copy on home page

petrjasek:
readItems were not populated on store


akintolga:
This reminded me `version` field in agenda was missing which I have added just now in my PR

PR #408, opened 2018-07-17
Disabled indexing on planning_items and event in agenda

akintolga:
- Also few bug fixes


newsroom/agenda/agenda.py
```diff
@@ -83,13 +83,10 @@ class AgendaResource(newsroom.Resource):
     schema['location'] = events_schema['location']
 
     # event details
-    schema['event'] = events_schema
+    schema['event'] = not_enabled
```
petrjasek: it should be like:
```
    # event details
    schema['event'] = {
        'type': 'object',
        'mapping': not_enabled,
    }

    # planning details which can be more than one per event
    schema['planning_items'] = {
        'type': 'list',
        'mapping': not_enabled,
    }
```
otherwise we don't push any mapping for it

PR #412, opened 2018-07-19
Place filter dropdown

akintolga:


assets/agenda/components/AgendaFilters.jsx
```diff
@@ -8,6 +8,9 @@ const filters = [{
 }, {
     label: 'Any location',
     field: 'location'
+}, {
+    label: 'Any region',
```
petrjasek: the labels here can use some gettext

akintolga: I used `gettext` doen below here: https://github.com/akintolga/newsroom/blob/d158dce4a742b7e7454616f5f2ed56fa0898c5b5/assets/agenda/components/AgendaFilters.jsx#L42

petrjasek: yep it should be there too, but it is needed here as well for string extraction

PR #416, opened 2018-07-23
[SDAN-264] - Account created email

akintolga:
- A new account created email is sent when a new user is created. This email will contain a reset password link
- Reset password button will trigger a reset password email


petrjasek:
btw this could be done on master and we can merge master to feat-agenda later

petrjasek:
pls make a PR for master

PR #428, opened 2018-08-10
[SDAN-288] - Picture coverage icon not displayed

akintolga:


akintolga:
I think the coverage types are set in vocabularies: https://github.com/superdesk/superdesk-planning/blob/master/server/data/vocabularies.json#L9 so I added a `picture` in addition to photo in css to cover that

PR #435, opened 2018-08-14
[SDAN-249] - Coverage request email

akintolga:
Backend changes


newsroom/agenda/views.py
```diff
@@ -8,8 +8,9 @@
 from newsroom.navigations.navigations import get_navigations_by_company
 from flask import current_app as app
 from newsroom.auth import get_user, login_required
-from newsroom.utils import get_entity_or_404, is_json_request
+from newsroom.utils import get_entity_or_404, is_json_request, get_json_or_400
 from newsroom.wire.views import update_action_list
+from newsroom.email import send_coverage_request_email
```
petrjasek: I'm not sure why we should define module specific emails in `newsroom.email`, imo it should be in `agenda.emails`

PR #440, opened 2018-08-16
[SDAN-249] - Request coverage ui

akintolga:


assets/agenda/components/AgendaCoverageRequest.jsx
```diff
@@ -0,0 +1,80 @@
...
+        this.setState({opened: false, message: '', status: null});
+    }
+
+    onMessageChange(event) {
+        this.setState({message: event.target.value, status: null});
+    }
+
+    requestCoverage() {
+        this.props.requestCoverage(this.props.item, this.state.message)
+            .then(() => this.setState({'status': gettext('Your request has been sent successfully')}))
```
petrjasek: I think this should go to action and use the notify service to be consistent

petrjasek: and it should probably reset after

PR #453, opened 2018-08-24
handle killed agenda as canceled

petrjasek:
and fix css class name


assets/agenda/components/AgendaPreview.jsx
```diff
@@ -39,7 +39,7 @@ class AgendaPreview extends React.PureComponent {
             'wire-column__preview--covering': hasCoverages(item),
             'wire-column__preview--not-covering': !hasCoverages(item),
             'wire-column__preview--postponed': isPostponed(item),
-            'wire-column__preview--cancelled': isCanceled(item),
```
akintolga: @petrjasek I think planning may have the state as `cancelled`: https://github.com/superdesk/superdesk-planning/blob/60d86732d7653afe26a2f3f9e340b55ef26431e4/server/planning/common.py#L39

petrjasek: it might, but we don't have cancelled in css, only canceled

PR #454, opened 2018-08-24
fix date filtering

petrjasek:
- when selecting a date show events which have any overlap with that day
- display each event in all groups it should be in, not just first

### DIFF
--- a/newsroom/agenda/agenda.py
+++ b/newsroom/agenda/agenda.py
@@ -162,13 +162,12 @@ def _agenda_query():
 
 
 def _event_date_range(args):
-    _range = {}
     offset = int(args.get('timezone_offset', '0'))
     if args.get('date_from'):
-        _range['gte'] = get_local_date(args['date_from'], '00:00:00', offset)
-    if args.get('date_to'):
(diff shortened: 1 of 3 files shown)


newsroom/agenda/agenda.py
```diff
@@ -162,13 +162,12 @@ def _agenda_query():
...
 def _event_date_range(args):
-    _range = {}
     offset = int(args.get('timezone_offset', '0'))
     if args.get('date_from'):
-        _range['gte'] = get_local_date(args['date_from'], '00:00:00', offset)
-    if args.get('date_to'):
-        _range['lte'] = get_local_date(args['date_to'], '23:59:59', offset)
-    return {'range': {'dates.start': _range}}
+        return [  # Each range query must be on its own, otherwise elastic doesn't return what it should.
+            {'range': {'dates.start': {'lte': get_local_date(args['date_from'], '23:59:59', offset)}}},
```
akintolga: This looks like creating a range for the first day of the argument `date_from`, is that right? And just double checking if we needed a `date_to` (in case we have a date range filter in navigation)

petrjasek: yep this is working for that top bar date picker, seems like it only produces single date

PR #455, opened 2018-08-26
[SDAN-34] - Location typeahead

akintolga:


petrjasek:
there is some duplication in the components, we should avoid it

akintolga:
Did further refactoring

PR #457, opened 2018-08-27
fix pushing planning for existing event

petrjasek:
keep event data as main, only add what's missing from planning item

### DIFF
--- a/newsroom/push.py
+++ b/newsroom/push.py
@@ -293,30 +293,48 @@ def set_agenda_metadata_from_event(agenda, event):
 
 
 def format_qcode_items(items=None):
-    if items:
+    try:
         return [{'code': item.get('qcode'), 'name': item.get('name')} for item in items]
+    except (TypeError, AttributeError):
+        return []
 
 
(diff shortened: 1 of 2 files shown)


newsroom/push.py
```diff
@@ -293,30 +293,48 @@ def set_agenda_metadata_from_event(agenda, event):
...
+    agenda['headline'] = get('headline')
+    agenda['slugline'] = get('slugline')
+    agenda['abstract'] = get('abstract')
+    agenda['ednote'] = get('ednote')
+    agenda['place'] = get('place')
+    agenda['subject'] = get('subject')
+    agenda['genre'] = planning_item.get('genre') or agenda.get('genre')
+    agenda['priority'] = planning_item.get('priority') or agenda.get('priority')
+    agenda['urgency'] = planning_item.get('urgency') or agenda.get('urgency')
+    agenda['products'] = event.get('products') or planning_item.get('products') or agenda.get('products')
```
akintolga: this could be `get(products)`

PR #463, opened 2018-08-28
[SDAN-290] - Added date into copy function

akintolga:


mdhaman:
@akintolga we having console error on `copy` action. I think it could be related to `abstract` missing for auto published items.

https://github.com/superdesk/newsroom/blob/b44adcb5fea0bdac6172a89687deedbed960957b/assets/wire/actions.js#L113

PR #477, opened 2018-09-10
[SDAN-298] - Ordering list of event in a day

akintolga:


petrjasek:
wouldn't this break the predefined sorting for featured?

akintolga:
Waiting a decision from AAP regarding how the `full day` events will be sorted

PR #482, opened 2018-09-16
[SDAN-306] - Displaying events for coverage dates

akintolga:
- Added a new field `extra_dates` to agenda schema for coverage or planning dates that falls outside of event start and end dates
- Changed agenda search to include that field as well
- When grouping agenda items on client side those extra days are also taken into account
- Hiding coverages for event days where coverage is on different day


petrjasek:
I would call it `display_dates` or something, extra doesn't give much info what it's for. also the `start` in the schema is implicating there could be and `end`

akintolga:
Updated the field

PR #490, opened 2018-09-20
Bootstrap update to v4.1.3

darconny:


petrjasek:
do we have to update some components and add that data-display too?

darconny:
> do we have to update some components and add that data-display too?

@petrjasek On static pages dropdown position was wrong and data-display parameter fixes it. But in nrpr-490.test.superdesk.org/wire everything looks ok. Although we can't test Agenda view (no content there), it should be the same as in wire.

PR #491, opened 2018-09-22
[SDAN-309] - Find users by email case insensitive

akintolga:


newsroom/auth/__init__.py
```diff
@@ -39,4 +40,20 @@ def get_user_id():
...
+    return _get_user_by_email(email, 'auth_user')
+
+
+def get_user_by_email(email):
+    """ Returns the user from users by the email case insensitive """
+    return _get_user_by_email(email, 'users')
+
+
+def _get_user_by_email(email, repo):
+    lookup = {'email': {'$regex': re.compile('^{}$'.format(email), re.IGNORECASE)}}
```
petrjasek: the email should be escaped imo

PR #509, opened 2018-10-04
Cached results doesn't reflect changes for bookmarks

akintolga:


petrjasek:
this will prevent any caching for searches, just wondering if we shouldn't rather use some if-modified-since header or something

akintolga:
I see, I'll check it how can we use if-modified-since here. It's the same for agenda search as well. I'm merging like this for now but will follow up on this.

PR #511, opened 2018-10-05
ignore settings when running tests

petrjasek:


petrjasek:
cherry picking from stt branch

PR #512, opened 2018-10-07
Notifications for agenda updates

akintolga:


assets/components/NotificationListItem.jsx
```diff
@@ -6,7 +6,7 @@ import CloseButton from './CloseButton';
 
 
 const getAgendaNotification = (item) => [
-    <div key='message' className="notif__list__info">{gettext('An agenda you downloaded has been updated')}</div>,
+    <div key='message' className="notif__list__info">{gettext('An agenda in your watch has been updated')}</div>,
```
petrjasek: not sure about the english, shouldn't it be `in your watch list` or something?

akintolga: You're probably right, again I merge it like this for now, but double check it with Karel

PR #517, opened 2018-10-13
[SDAN-329] - Agenda filters

akintolga:
- Search the agenda by date range
  - Today, This week and This Month should only return results for that that range
- Display the number of results only when the query takes place
- Query only when user clicks search button

### DIFF
--- a/assets/agenda/actions.js
+++ b/assets/agenda/actions.js
@@ -128,6 +128,11 @@ function search(state, next) {
     const createdFilter = get(state, 'search.createdFilter', {});
     const agendaDate = getDateInputDate(get(state, 'agenda.activeDate'));
 
+    let dateTo = createdFilter.to;
+    if (createdFilter.from && createdFilter.from.indexOf('now') >= 0) {
+        dateTo = createdFilter.from;
+    }
+
     const params = {
         q: state.query,
         id: state.queryId,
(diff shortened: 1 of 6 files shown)


assets/agenda/actions.js
```diff
@@ -128,16 +128,20 @@ function search(state, next) {
     const createdFilter = get(state, 'search.createdFilter', {});
     const agendaDate = getDateInputDate(get(state, 'agenda.activeDate'));
 
+    let dateTo = createdFilter.to;
+    if (createdFilter.from && createdFilter.from.indexOf('now') >= 0) {
```
petrjasek: you can use `createdFilter.from.includes('now')` instead of `indexOf`

assets/reducers.js
```diff
@@ -133,8 +133,10 @@ export function defaultReducer(state, action) {
         return {...state, query: action.query, activeItem: null, search: search};
     }
 
-    case QUERY_ITEMS:
-        return {...state, isLoading: true, totalItems: null, activeQuery: state.query};
+    case QUERY_ITEMS: {
+        const resultsFiltered = !isEmpty(get(state, 'search.activeFilter')) || !isEmpty(get(state, 'search.createdFilter'));
```
petrjasek: I'm not sure if we should keep on state something derived from state, that's kind of breaking this single source of truth idea, we should probably have selector defined for it and reuse that if we need it in more places

akintolga: I think we need to understand if a query took place. User can set a filter or change the date range, so `search.activeFilter` or `search.createdFilter` assigned some values. But that doesn't mean the results are filtered yet until they click the `search` button. That's why I moved this line from selector to the action.

akintolga:
I have run this PR with AAP, they want removal of the filters to trigger `search` action automatically. So when a new filter is selected user needs to click `search` button but when a filter is removed search will trigger automatically.

akintolga:
Closing this as AAP said they are happy with the autosearch PR#527

PR #524, opened 2018-10-16
[SDAN-325] - Map move

akintolga:


assets/agenda/components/AgendaItemDetails.jsx
```diff
@@ -50,10 +54,8 @@ export default function AgendaItemDetails({item, user, actions, onClose}) {
                 </ContentBar>
             </ContentHeader>
 
-            <Article image={map} covering={hasCoverages(item)}>
+            <Article covering={hasCoverages(item)} header={header}>
```
petrjasek: might be cleaner to pass there item and handle the map name and time in article component

PR #528, opened 2018-10-19
implement general settings with google analytics config

petrjasek:
STTNHUB-10

### DIFF
--- a/assets/general-settings/actions.js
+++ b/assets/general-settings/actions.js
@@ -1,0 +1,25 @@
+import server from 'server';
+import { gettext, notify } from 'utils';
+
+export const UPDATE_VALUES = 'UPDATE_VALUES';
+function updateValues(values) {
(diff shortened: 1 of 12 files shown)


assets/general-settings/actions.js
```diff
@@ -0,0 +1,25 @@
...
+
+export function save(values) {
+    return (dispatch) => {
+        server.post('/settings/general_settings', values)
+            .then(() => {
+                notify.success(gettext('Settings were updated successfully.'));
+                dispatch(updateValues(values));
+            }, (reason) => {
+                console.error('error', reason);
+                notify.error(gettext('Ups, there was an error when updating settings.'));
```
akintolga: I'd rather say `Oops` here :)

petrjasek: right, there must be a trademark for it :wink:

PR #531, opened 2018-10-20
[SDAN-318] - Coverage tooltips

akintolga:
Also addresses [SDAN-319]

### DIFF
--- a/assets/agenda/components/AgendaListItemIcons.jsx
+++ b/assets/agenda/components/AgendaListItemIcons.jsx
@@ -7,7 +7,8 @@ import {
     hasLocation,
     getLocationString,
     getCoverageIcon,
-    isRecurring
+    isRecurring,
+    WORKFLOW_STATUS_TEXTS
 } from '../utils';
 
 import AgendaListItemLabels from './AgendaListItemLabels';
@@ -22,17 +23,10 @@ function AgendaListItemIcons({item, group, hideCoverages, row}) {
 
     const getCoverageTootip = (coverage) => {
 
-        const status = {
-            draft: 'planned',
-            assigned: 'planned',
-            active: 'in progress',
-            completed: 'available',
-        };
-
         if (['draft', 'assigned', 'active'].includes(coverage.workflow_status)) {
             return gettext('{{ type }} coverage {{ status }}, due {{date}} at {{time}}', {
                 type: coverage.coverage_type,
-                status: status[coverage.workflow_status],
+                status: WORKFLOW_STATUS_TEXTS[coverage.workflow_status],
                 date: formatDate(coverage.scheduled),
                 time: formatTime(coverage.scheduled)
             });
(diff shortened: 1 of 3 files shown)


assets/agenda/utils.js
```diff
@@ -32,6 +32,14 @@ const Groupers = {
     'month': formatMonth,
 };
 
+export const WORKFLOW_STATUS_TEXTS = {
```
petrjasek: this can use some gettext annotations

petrjasek: this can use some gettext

petrjasek: this can use some gettext

assets/agenda/components/CoverageItemStatus.jsx
```diff
@@ -35,10 +36,11 @@ export default class CoverageItemStatus extends React.PureComponent {
     render() {
         const {coverage} = this.props;
 
-        if (coverage.coverage_status && !this.state.wire) {
+        if (coverage.workflow_status && !this.state.wire) {
             return [
                 <span key="label" className='coverage-item__text-label mr-1'>{gettext('Status')}:</span>,
-                <span key="value">{coverage.coverage_status}</span>,
+                <span key="value">{gettext('coverage {{ state }} ',
+                    {state: get(WORKFLOW_STATUS_TEXTS, coverage.workflow_status, 'planned')})}</span>,
```
akintolga: Here is another translation

assets/agenda/components/AgendaListItemIcons.jsx
```diff
@@ -22,17 +23,10 @@ function AgendaListItemIcons({item, group, hideCoverages, row}) {
...
-            assigned: 'planned',
-            active: 'in progress',
-            completed: 'available',
-        };
-
         if (['draft', 'assigned', 'active'].includes(coverage.workflow_status)) {
             return gettext('{{ type }} coverage {{ status }}, due {{date}} at {{time}}', {
                 type: coverage.coverage_type,
-                status: status[coverage.workflow_status],
+                status: WORKFLOW_STATUS_TEXTS[coverage.workflow_status],
```
akintolga: Here is the first usage where its is getting translated

akintolga:
Hi @petrjasek I did the translations two places it is used as above. Should I still use it in the object itself as well?

assets/agenda/components/CoverageItemStatus.jsx
petrjasek: here it will translate `coverage {{ state }}` and replace state with workflow status, but that must be translated on its own

petrjasek:
yep you should add it to the object for gettext extract to pick it up

assets/agenda/components/CoverageItemStatus.jsx
akintolga: I see, cool I'll do that

akintolga:
Added gettext

PR #544, opened 2018-10-26
Stt theme

darconny:


newsroom/default_settings.py
```diff
@@ -77,8 +77,8 @@
     'newsroom.agenda'
 ]
 
-SITE_NAME = 'AAP Newsroom'
-COPYRIGHT_HOLDER = 'AAP'
+SITE_NAME = 'Newsroom'
```
petrjasek: this shouldn't be needed, it's set already in settings.py
and it breaks tests..

PR #545, opened 2018-10-28
[SDAN-321] - Picture coverage link

akintolga:
Creating a placeholder module for generating delivery href for coverages

### DIFF
--- a/newsroom/photos.py
+++ b/newsroom/photos.py
@@ -1,0 +1,6 @@
+def set_photo_coverage_href(coverage, planning_item):
+    pass
+
+
+def init_app(app):
+    app.set_photo_coverage_href = set_photo_coverage_href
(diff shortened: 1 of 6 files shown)


newsroom/push.py
```diff
@@ -413,6 +413,14 @@ def should_add(date):
     return display_dates
 
 
+def set_coverage_delivery_href(coverage, slugline):
```
petrjasek: I think we will need some api to define the function for delivery href rather than using just photo url config and adding some hardcoded params, maybe defining `app.delivery_href` function which be noop by default and could be set in some app specific delivery module

akintolga: Cool, I wasn't sure if this was the right way. Discussde with @mdhaman I'll change the PR to have an API and noop

newsroom/photos.py
```diff
@@ -0,0 +1,6 @@
+def set_photo_coverage_href(coverage, slugline):
```
petrjasek: imo we should pass there `coverage` and `planning_item`, not the `slugline` which is already kind of aap specific

PR #553, opened 2018-10-31
[SDAN-308](feat): Adding section for content market place.

mdhaman:
- Add new section "Market Place".
- UI Config for Preview and Item Details.
- Images can be added to navigations so that card display can be created.
- Removed hardcoded bookmarks URL from BookmarksTab component.


newsroom/ui_config.py
```diff
@@ -0,0 +1,51 @@
+import superdesk
+import newsroom
+
+
+class UIConfigResource(newsroom.Resource):
```
petrjasek: any reason to store this in db? just wondering

PR #563, opened 2018-11-08
[SDAN-345] - Linking wire item from ingested coverage

akintolga:
This covers the cases where coverage contains the link to an already published wire story

### DIFF
--- a/newsroom/push.py
+++ b/newsroom/push.py
@@ -7,7 +7,7 @@ from datetime import datetime
 
 from copy import copy, deepcopy
 from PIL import Image, ImageEnhance
-from flask import current_app as app
+from flask import current_app as app, url_for
 from flask_babel import gettext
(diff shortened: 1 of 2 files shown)


newsroom/push.py
```diff
@@ -421,6 +421,11 @@ def get_coverages(planning_items, original_coverages=[]):
     def get_existing_coverage(id):
         return next((o for o in original_coverages if o['coverage_id'] == id), {})
 
+    def set_text_delivery(coverage, deliveries):
+        if coverage['coverage_type'] == 'text' and deliveries:
+            coverage['delivery_id'] = deliveries[0]
```
akintolga: @mdhaman This should be like `deliveries[0]['item_id']` is that right?

mdhaman: yes

PR #584, opened 2018-11-24
[SDAN-355] - Fix for the agenda search dates

akintolga:
- Removed Content Type
- Changed the date calculation for agenda to start the search date from time:'00:00:00' of the local date of the user

### DIFF
--- a/newsroom/agenda/agenda.py
+++ b/newsroom/agenda/agenda.py
@@ -1,9 +1,6 @@
-import pytz
 import logging
-from datetime import datetime, timedelta
 
 from content_api.items.resource import code_mapping
-from dateutil.relativedelta import relativedelta
 from eve.utils import ParsedRequest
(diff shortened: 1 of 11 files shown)


newsroom/agenda/agenda.py
```diff
@@ -207,6 +206,35 @@ def _agenda_query():
     }
 
 
+def get_utcnow():
+    'added for unit tests'
```
petrjasek: this should be probably `'''...'''`

newsroom/agenda/agenda.py
```diff
@@ -207,6 +206,35 @@ def _agenda_query():
...
+    return datetime.utcnow()
+
+
+def today(time, offset):
```
petrjasek: imo would be good to put those date related functions into some date_utils module, so we can reuse it and it's not agenda related really

PR #592, opened 2018-11-26
Fix for download and copy actions

akintolga:
[SDAN - 349] - Error on download
[SDAN - 350] - Copy action was missing in agenda
[SDAN - 368] - Redundant open action in opened item


assets/wire/actions.js
```diff
@@ -37,9 +38,9 @@ export function preview(item) {
 }
 
 export function previewAndCopy(item) {
-    return (dispatch) => {
+    return (dispatch, getState) => {
         dispatch(previewItem(item));
-        dispatch(copyPreviewContents(item));
+        dispatch(getState().context === 'agenda' ? copyPreviewContentsOfAgendaItem(item) : copyPreviewContents(item));
```
petrjasek: I guess agenda should call different action and not something from wire

akintolga: Yes, In that case I need to refactor this bit to pass in different action, what do you think? https://github.com/superdesk/newsroom/blob/master/assets/wire/item-actions.js#L10

petrjasek: yep imo the actions should be constructed in the main app component, wireApp or agendaApp

PR #604, opened 2018-11-28
[SDAN-376] Adding settings to override templates for emails and print.

mdhaman:


petrjasek:
you can override it already in theme folder, do you need this to be configurable via env var?

mdhaman:
I will use the theme folder to override templates. closing this PR.

PR #613, opened 2018-11-30
STTNHUB-39

darconny:
@petrjasek  layout.html is just static code, please add that to the js logic.


petrjasek:
it's merged on master

PR #623, opened 2018-12-05
make aggregations and filter groups configurable

petrjasek:
STTNHUB-3


newsroom/push.py
```diff
@@ -126,6 +127,7 @@ def publish_item(doc):
         agenda_items = superdesk.get_resource_service('agenda').set_delivery(doc)
         if agenda_items:
             [notify_new_item(item, check_topics=False) for item in agenda_items]
+    publish_item_signal.send(app._get_current_object(), item=doc)
```
akintolga: Just curious, it looks like there's no listener for this for now, it that right?

petrjasek: yep, using it on stt branch

PR #637, opened 2018-12-11
[SDAN-362] - Additional metadata for planning items

akintolga:


assets/agenda/utils.js
```diff
@@ -405,6 +405,25 @@ export function getInternalNotesFromCoverages(item) {
     return internalNotes;
 }
 
+/**
+ * Get list of subjects
+ *
+ * @param {Object} item
+ * @return {Array}
+ */
+export function getSubjects(item) {
```
petrjasek: shouldn't this check for duplicates at some point?

akintolga: True. I'll add that check.

akintolga:
I have added the uniqe check

assets/agenda/utils.js
```diff
@@ -405,6 +405,25 @@ export function getInternalNotesFromCoverages(item) {
...
+    let subjects = get(item, 'event.subject') || [];
+    const planningItems = get(item, 'planning_items', []);
+    (planningItems || []).forEach(p => {
+        subjects = subjects.concat(p.subject || []);
+        get(p, 'coverages', []).forEach(c => {
+            subjects = subjects.concat(c.subject || []);
+        });
+    });
+
+    return uniqBy(subjects.filter(subject => !isEmpty(subject)), 'qcode');
```
petrjasek: I think in newsroom we store subjects with `code` not `qcode`

akintolga: Actually, in wire we missed that coversion. Also by checking the `push` again I just realized that we don't need to check every planning item. So I'll change it again.

akintolga:
Updated `push` to handle subject over there

PR #640, opened 2018-12-16
[SDAN-397] - Adding filter parameters into url

akintolga:
- With this change selected filters and navigation is reflected in URL
- On reload of the page all those values are applied 
- Now when sharing a topic the url will have all the parameters of the topic

### DIFF
--- a/assets/wire/actions.js
+++ b/assets/wire/actions.js
@@ -11,6 +11,7 @@ import {
     toggleNavigation,
     setCreatedFilter,
     toggleNavigationById,
+    initParams as initSearchParams,
 } from 'search/actions';
 
 
@@ -212,6 +213,9 @@ export function fetchItems() {
                 const state = getState();
                 updateRouteParams({
                     q: state.query,
+                    filter: get(state, 'search.activeFilter'),
+                    navigation: get(state, 'search.activeNavigation'),
+                    created: get(state, 'search.createdFilter'),
                 }, state);
                 analytics.timingComplete('search', Date.now() - start);
             })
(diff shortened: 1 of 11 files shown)


newsroom/topics/views.py
```diff
@@ -54,6 +54,26 @@ def is_user_topic(topic_id, user_id):
     return False
 
 
+def get_topic_url(topic):
```
petrjasek: wondering if we should rather use `topic=<topic-id>` and handle it on server, if topics are editable the url would have to change, but if it was shared already it won't work

akintolga: We could use id too. Afaik user cannot edit a topic (except name) but they can delete and recreate of course. Then shared email won't be valid. I was inclined to use URL as one feedback from AAP few weeks back was to use URL for the dropdown filters in agenda (which I did local store).

petrjasek: I see, if the query is not editable it's ok

assets/wire/actions.js
```diff
@@ -212,6 +213,9 @@ export function fetchItems() {
                 const state = getState();
                 updateRouteParams({
                     q: state.query,
+                    filter: get(state, 'search.activeFilter'),
```
mdhaman: Different browser have different limits on saving data to history. `updateRouteParams` fails on firefox, ie, etc after hitting the limit. 
You would see data changing as dispatch for updating the store is before `updateRouteParams` but the URL stops changing.

akintolga: Hi @mdhaman I think the problem is the sixe of the state being too big for firefox to handle so as you say it was hitting the limit. I made the change now to trim the `items` and `itemsById` which were the biggest chunk in state. Now we fetch once the hidtory is popped back.

PR #644, opened 2018-12-19
[SDAN-384] - Make list item mouseover animation configurable

akintolga:


newsroom/default_settings.py
```diff
@@ -205,6 +205,9 @@
 # Hides or displays abstract on preview panel and details modal
 DISPLAY_ABSTRACT = False
 
+# Enables or disables the animations for list item select boxes
+USE_ANIMATIONS = True
```
petrjasek: imo could be called `LIST_ANIMATIONS`, in case there would be any other, but maybe we can just have some `CLIENT_CONFIG` dict which we would send to client, so we don't have to add new settings for every toggle

PR #655, opened 2019-01-04
[SDAN-387] - Layout update for agenda print

akintolga:


newsroom/agenda/views.py
```diff
@@ -112,3 +123,13 @@ def follow():
     update_action_list(data.get('items'), 'watches', item_type='agenda')
     push_user_notification('saved_items', count=get_resource_service('agenda').get_saved_items_count())
     return flask.jsonify(), 200
+
+
+@blueprint.route('/agenda_print', methods=['POST'])
+@login_required
+def print_agenda():
```
petrjasek: how is this useful?

akintolga: The new shape of the print template has fields like event date&time, contacts, location details and the map which we have the business rules on client side. Instead of creating the same rules for the backend, I wanted to pass that data to backend.

petrjasek: I see, imo it would be better then to render the print page in response to that post and avoid the cache

akintolga:
Hi @petrjasek , I created helper functions except the map url which I am still passing from client. Please let me know what you think.

PR #668, opened 2019-01-17
[STTNHUB-48] - Character count option for stories

akintolga:


assets/utils.js
```diff
@@ -332,13 +332,50 @@ export function getTextFromHtml(html) {
     return text.join('');
 }
 
+
+/**
+ * Get label for the count field
+ */
+export function getCountLabel(titleCase) {
+    const text = gettext(getConfig('count_label', 'words'));
```
petrjasek: this won't extract the label for translation

petrjasek:
btw according to the ticket they want to have both word and char count, not just one

akintolga:
Hi @petrjasek you're right, I'll add character count as an extra. Thanks for the notification 👍

assets/wire/components/WireListItem.jsx
```diff
@@ -98,7 +98,7 @@ class WireListItem extends React.Component {
                                 <div className='wire-articles__item__meta-info'>
                                     <span className='bold'>{this.slugline}</span>
                                     <span>{item.source}
-                                        {' // '}<span>{this.wordCount}</span> {gettext('words')}
+                                        {' // '}<span>{this.count}</span> {gettext('words')}
```
nrvikas: The variable (this.count) needs initialization (or assignment) - as this count value is not seen on the wire's list item.

PR #673, opened 2019-01-22
[SDAN-413] Display internal note for internal users in agenda's list, preview, read views

nrvikas:


petrjasek:
just wondering how sensitive this info can be, but imo it would be better to handle it on backend - if the user is not internal type don't return internal note, and on client just display it if it's there

akintolga:
@nrvikas Fyi there are few backend pieces that you may want to check if they work for you:
This is for not fetching the sensitive items from elastic: https://github.com/superdesk/newsroom/blob/master/newsroom/agenda/agenda.py#L342

This is when single item is fetched removing before returning: https://github.com/superdesk/newsroom/blob/master/newsroom/agenda/views.py#L43

nrvikas:
Good idea, thanks.

PR #682, opened 2019-01-24
[SDAN-426] 'My Topics' to be seen only in 'wire' context

nrvikas:


nrvikas:
Added a 'WIP' to see if the user has no permission for 'wire', my topics should not appear anywhere in the site (ex. under the drop-list of user-avatar).

PR #686, opened 2019-01-25
[SDAN-428] Applying styles to google maps.

mdhaman:


newsroom/settings.py
```diff
@@ -48,9 +50,20 @@ def get_setting(setting_key=None):
     return flask.g.settings
 
 
+def get_client_config():
+    config = newsroom_config()
+    keys = ['google_maps_styles']
```
petrjasek: imo would be good to add some switch to `app.general_setting` api instead of hardcoding the keys here

PR #687, opened 2019-01-29
[SDAN-430] Point url to section in history/bookmark notification emails

MarkLark86:
* Store section with history entries
* wire/search: get_matching_bookmarks should apply section/product filters

### DIFF
--- a/newsroom/wire/search.py
+++ b/newsroom/wire/search.py
@@ -444,7 +444,22 @@ class WireSearchService(newsroom.Service):
         """
         bookmark_users = []
 
-        search_results = self.get_items(item_ids)
+        query = {
+            'bool': {
+                'must_not': [
+                    {'term': {'type': 'composite'}},
+                ],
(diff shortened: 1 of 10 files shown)


newsroom/wire/search.py
```diff
@@ -444,7 +444,30 @@ def get_matching_bookmarks(self, item_ids, active_users, active_companies):
...
+            'bool': {
+                'must_not': [
+                    {'term': {'type': 'composite'}},
+                ],
+                'must': [
+                    {'terms': {'_id': item_ids}}
+                ],
+            }
+        }
+        user = get_user()
```
mdhaman: `get_matching_bookmarks` is called from `notify_user_matches` in push.py. There won't be any `user` or `company`.  We don't need `set_product_query` or `set_bookmarks_query` functions as well.

MarkLark86: Done

PR #700, opened 2019-02-12
[SDAN-274] Celery task and manage-command to send company expiry alerts

nrvikas:


newsroom/company_expiry_alerts.py
```diff
@@ -0,0 +1,65 @@
...
+                'expiry_date': expiry_time
+            }
+
+            with app.mail.connect() as connection:
+                send_email(
+                    recipients,
+                    'Companies expiring within next 7 days ({})'.format(expiry_time.strftime('%Y-%m-%d')),
+                    text_body=render_template('company_expiry_email.txt', **template_kwargs),
+                    html_body=render_template('company_expiry_email.html', **template_kwargs),
+                    sender='vnarayanaswamy@aap.com.au',
```
mdhaman: We don't need sender. It is already configured see: 

https://github.com/superdesk/newsroom/blob/master/newsroom/default_settings.py#L133
https://github.com/superdesk/newsroom/blob/master/newsroom/email.py#L21

Do we need `connection` or `send_email` function handles it?

newsroom/flaskapp.py
```diff
@@ -193,6 +199,11 @@ def _setup_theme(self):
             view_func=self.send_theme_file
         )
 
+    def _setup_celery_serialization(self):
```
mdhaman: Do we need this if we register the serializer with different name i.e. `newsroom/json`?

newsroom/templates/company_expiry_email.html
```diff
@@ -0,0 +1,7 @@
+<h3>The following company accounts are due to expire within next 7 days (before {{expiry_date.strftime('%Y-%m-%d')}}):</h3>
+
+<ul>
+    {% for company in companies %}
+        <h3><pre>&nbsp;{{ company.get('name') }} ({{ company.get('contact_email') }}), Tel: ({{ company.phone }}) expiring on {{ company.expiry_date.strftime('%Y-%m-%d')}}</pre></h3>
```
mdhaman: Check for the existence of `contact_email` and `phone` in the template. Otherwise it will render email as `AAP (), Tel: () expiring in 25/02/2019`.

worker.py
```diff
@@ -0,0 +1,19 @@
...
+# at https://www.sourcefabric.org/superdesk/license
+
+
+import logging
+from newsroom.flaskapp import Newsroom
+
+
+logger = logging.getLogger(__name__)
+app = Newsroom("newsroom_celery")
+celery = app.celery
```
petrjasek: it would be better to have this file in newsroom module, so you can use it in newsroom-aap without making a copy of it

newsroom/celery_app.py
```diff
@@ -0,0 +1,166 @@
...
+            handle_exception(exc)
+
+
+celery.Task = AppContextTask
+
+
+def init_celery(app):
+    celery.config_from_object(app.config, namespace='CELERY')
+    app.celery = celery
+    app.redis = __get_redis(app)
```
petrjasek: I think we don't need this whole subtask progress part here

petrjasek: I mean from this line to the end it can be removed. actually wondering why not import the rest from superdesk core?

newsroom/company_expiry_alerts.py
nrvikas: That sender name is by mistake totally, sorry about that.

newsroom/templates/company_expiry_email.html
nrvikas: I thought of keeping it this way so that the administrators can be alerted that the company is missing these details. Can we use (-) instead of () ? Or check for existence and not show if it doesn't exist ?

mdhaman: or change the template to AAP (Email: - , Tel: -) expiring on <date>

newsroom/flaskapp.py
nrvikas: Nope, I will remove it.

nrvikas:
@petrjasek and @mdhaman , thanks for the suggestions. The suggested changes are done.

PR #701, opened 2019-02-12
[SDAN-435] Using 16:9 image renditions for preview and details.

mdhaman:


assets/wire/utils.js
```diff
@@ -67,8 +68,26 @@ export function getThumbnailRendition(picture, large) {
  * @param {Object} picture
  * @return {Object}
  */
-export function getPreviewRendition(picture) {
-    return get(picture, 'renditions._newsroom_view', get(picture, 'renditions.viewImage'));
+export function getPreviewRendition(picture, isCustom = false) {
+    return get(
+        picture,
+        isCustom ? 'renditions._newsroom_custom' : 'renditions._newsroom_base',
```
petrjasek: just wondering - wouldn't it make sense to change that `_newsroom_base` rendition based on some config instead of adding this logic to client? or we still use that `_newsroom_base` somewhere when `_newsroom_custom` is defined?

mdhaman: Image renditions for existing stories is not changed. older stories with are still going to use `_newsroom_base` renditions.

petrjasek: I see

PR #707, opened 2019-02-15
In wire, scroll bars was going to top when filters were used and an item was previewed

nrvikas:


nrvikas:
@petrjasek I have updated this PR with more changes if you want to re-review it.

PR #718, opened 2019-02-25
[SDAN-460][SDAN-461] Coverage schedule display changes

nrvikas:


newsroom/agenda/agenda.py
```diff
@@ -444,13 +444,20 @@ def _enhance_items(self, docs):
...
             delivery_ids = [c.get('delivery_id') for c in (doc.get('coverages') or [])
                             if c['workflow_status'] == ASSIGNMENT_WORKFLOW_STATE.COMPLETED]
-            wire_items = get_resource_service('wire_search').get_items(delivery_ids)
+            wire_search_service = get_resource_service('wire_search')
+            wire_items = wire_search_service.get_items(delivery_ids)
             if wire_items.count() > 0:
+                _access = None
                 for item in wire_items:
+                    if _access is None:
+                        _access = wire_search_service.has_permissions(item, ignore_latest=True)
```
mdhaman: In this case, requesting permission `on demand` might be better as you would be doing no. of delivery ids + `get_item` query for all users. i.e. one page request could generate lot of request to elasticsearch queries. whereas in other case only for users who are preview or full details views. 

As per the PR, preview will be faster and search will be slower as it needs to calculate wire permissions on each request.

nrvikas: Yup, agreed. Will update the PR.

PR #743, opened 2019-03-22
Added TLS to default settings

akintolga:
- Also a potential fix for the boolean values provided

### DIFF
--- a/newsroom/default_settings.py
+++ b/newsroom/default_settings.py
@@ -1,5 +1,6 @@
 import os
 import tzlocal
+from distutils.util import strtobool as _strtobool
 
 from superdesk.default_settings import (   # noqa
     VERSION,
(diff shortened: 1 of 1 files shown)


newsroom/default_settings.py
```diff
@@ -117,7 +128,8 @@
 # Email settings
 MAIL_SERVER = os.environ.get('MAIL_SERVER') or 'localhost'
 MAIL_PORT = os.environ.get('MAIL_PORT') or 25
-MAIL_USE_SSL = os.environ.get('MAIL_USE_SSL') or False
+MAIL_USE_SSL = strtobool(os.environ.get('MAIL_USE_SSL')) or False
+MAIL_USE_TLS = strtobool(os.environ.get('MAIL_USE_TLS')) or False
```
petrjasek: I would rather import all these `MAIL*` settings from superdesk so we have it in sync

PR #744, opened 2019-03-27
[STTNHUB-54] - Displaying of the first published date

akintolga:
- I've been told that STT wanted to `firstcreated` field for date checks
- Now using `versioncreated` if a correction/kill arrives
- Added further translations


newsroom/push.py
```diff
@@ -76,7 +76,7 @@ def push():
         notify_new_item(agenda, check_topics=True)
     elif item.get('type') == 'text':
         orig = app.data.find_one('wire_search', req=None, _id=item['guid'])
-        item['_id'] = publish_item(item)
+        item['_id'] = publish_item(item, is_new=orig is None)
```
petrjasek: we should make that change on master and merge it here

akintolga: Not this PR depends on https://github.com/superdesk/newsroom/pull/745

PR #752, opened 2019-04-07
[STTNHUB-52] - Making list of languages parametric

akintolga:


newsroom/default_settings.py
```diff
@@ -250,8 +251,13 @@
     'list_animations': True,  # Enables or disables the animations for list item select boxes
 }
 
-LANGUAGES = ['en', 'fi', 'cs']
-DEFAULT_LANGUAGE = 'en'
+# List of languages available in login and profile settings
+# use string representation of a json i.e. '["en", "fi"]'
+try:
+    LANGUAGES = json.loads(os.environ.get('LANGUAGES'))
```
petrjasek: I don't think we need env for this kind of setting, this will be the same for stt in all environments, so we should just set it in their settings file

PR #761, opened 2019-04-16
[SDAN-403] Adding toggle filter for events only view.

mdhaman:


nrvikas:
Do we need to provide 'Events Only' filter option available when 'Top Stories' is active? It will never give any results when 'Events only' option is used.

nrvikas:
> Do we need to provide 'Events Only' filter option available when 'Top Stories' is active? It will never give any results when 'Events only' option is used.

Sorry, the above scenario works fine.

PR #763, opened 2019-04-18
Using agenda resource only for planning items

akintolga:


petrjasek:
is it any different?

newsroom/push.py
```diff
@@ -65,20 +65,19 @@ def push():
     item = flask.json.loads(flask.request.get_data())
     assert 'guid' in item or '_id' in item, {'guid': 1}
     assert 'type' in item, {'type': 1}
-    agenda_service = superdesk.get_resource_service('agenda')
```
akintolga: This line created an issue on `STT`, resource `agenda` is not available for `STT` but this gets executed even if we push only text items

PR #765, opened 2019-04-23
Displaying private notes if they exists

akintolga:


assets/wire/components/PreviewEdnote.jsx
```diff
@@ -1,13 +1,21 @@
...
+import { get } from 'lodash';
 import { gettext } from 'utils';
 import InfoBox from './InfoBox';
 
 
+/**
+ * @return {null}
+ */
 function PreviewEdnote({item, isItemDetail}) {
+    const privateNote = get(item, 'extra.sttnote_private');
```
petrjasek: I think it might be better to add sttnote_private contents to ednote field on push rather than using some custom stt field in ui

akintolga: If we do it in push we still have the custom code in push. I think the best would be to do this in the superdesk-core `ninjs_newsroom_formatter.py` or we extend this on superdesk-stt for a stt specific one. What do you think?

petrjasek: we have a signal on push so we can do it on stt branch in newroom like in
https://github.com/superdesk/newsroom/blob/stt/stt/filters.py#L11

akintolga: Cool, I'll that way

akintolga:
Closing this due to https://github.com/superdesk/newsroom/pull/766

PR #766, opened 2019-04-24
Adding private notes to ednote

akintolga:


stt/filters.py
```diff
@@ -25,6 +25,13 @@ def on_publish_item(app, item, is_new, **kwargs):
         if firstcreated < item['versioncreated']:
             item['versioncreated'] = firstcreated
 
+    # add private note to ednote
+    if item.get('extra', {}).get('sttnote_private'):
+        if item.get('ednote'):
+            item['ednote'] = '{} {}'.format(item['ednote'], item['extra']['sttnote_private'])
```
petrjasek: maybe we should use a line break there `{}\n{}`

akintolga: Would that render properly on UI?

petrjasek: I think it will render it with space, but we can use some replace `\n` with `<br>` later, if it's just space it won't be possible

PR #767, opened 2019-04-24
SDRLNE-4

darconny:


newsroom/default_settings.py
```diff
@@ -108,13 +108,13 @@
     'newsroom.media_utils',
 ]
 
-SITE_NAME = 'AAP Newsroom'
-COPYRIGHT_HOLDER = 'AAP'
+SITE_NAME = 'Медиасеть'
```
petrjasek: you shouldn't change those default_settings, but rather add `settings.py` file to root folder and put those changes there, like in https://github.com/superdesk/newsroom/blob/ansa/settings.py

darconny: Thanks please check now.

settings.py
```diff
@@ -0,0 +1,4 @@
+import os
```
petrjasek: you can remove this line, travis fails because of it

darconny:
@petrjasek Please check now and merge if its OK.

PR #778, opened 2019-05-16
[STTNHUB-65] - Embargoed items should not appear in cards

akintolga:


newsroom/wire/search.py
```diff
@@ -289,7 +289,11 @@ def get_product_items(self, product_id, size):
         if not product:
             return
 
-        query['bool']['should'] = []
+        query['bool']['should'] = [
```
petrjasek: I guess this should be part of must query otherwise if there is product query or product id both might be false and it will pass

PR #815, opened 2019-07-23
[SDAN-524][SDAN-530] Record 'open' and 'preview' actions in history collection

nrvikas:


newsroom/history.py
```diff
@@ -47,12 +58,13 @@ def create(self, docs, action, user, section='wire', **kwargs):
...
             return {
                 'action': action,
-                'created': now,
+                'versioncreated': now,
                 'user': user['_id'],
                 'company': user.get('company'),
                 'item': item['_id'],
                 'version': item.get('version', item.get('_current_version')),
                 'section': section,
+                'extra_data': item,
```
MarkLark86: For now, don't store the item in the history record. Leave the 'extra_data' in the schema for future use.
We will tackle this problem when we tackle purging of items from Newsroom.

PR #821, opened 2019-08-02
[SDAN-565] Reposition the 'show map' text in Agenda Preview

nrvikas:


MarkLark86:
Can you place `cursor: pointer` on the button for showing/hiding the map please

PR #832, opened 2019-08-22
[SDAN-578][SDAN-579] 'Account Manager' field in Company schema and use that in Company Expiry alerts

nrvikas:


MarkLark86:
You need to add `account_manager` to `SELECT_COMPANY` and `NEW_COMPANY` reducers so these values will default to a blank string for the `TextInput` field

nrvikas:
> You need to add `account_manager` to `SELECT_COMPANY` and `NEW_COMPANY` reducers so these values will default to a blank string for the `TextInput` field

Sure, thanks.

PR #866, opened 2019-10-15
[SDAN-552] Implement API token authentication

marwoodandrew:


newsroom/news_api/api_tokens/__init__.py
```diff
@@ -10,6 +13,32 @@
...
+class CompanyTokenAuth(TokenAuth):
+    def check_auth(self, token, allowed_roles, resource, method):
+        """Try to find auth token and if valid put subscriber id into ``g.user``."""
+        data = app.data.mongo.find_one(API_TOKENS, req=None, _id=token)
+        if not data:
+            return False
+        # Check if the token has expired
+        now = utcnow()
+        if data.get('expiry') < now:
+            app.data.mongo.remove(API_TOKENS, {'_id': token})
```
MarkLark86: For the News API, we won't delete tokens automatically if they expire (allowing an admin to update the expiry and the subscriber to keep using the same token).

PR #883, opened 2019-10-31
[SDAN-556][SDAN-557] News API return item in requested format

marwoodandrew:


newsroom/news_api/formatters/service.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def get_version(self, id, version, formatter_name):
+        formatter = self._get_formatter(formatter_name)
+        if not formatter:
+            abort(404)
+        if version:
+            item = get_resource_service('items_versions').find_one(req=None, _id_document=id, version=version)
+            resource_def = app.config['DOMAIN']['items']
+            id_field = versioned_id_field(resource_def)
+            item['_id'] = item[id_field]
```
MarkLark86: It is possible that `item` may be `None` here if `find_one` fails to find the item

PR #884, opened 2019-10-31
[SDAN-441] User and Watch List Administrator activities

nrvikas:


assets/search/components/WatchListEditor.jsx
```diff
@@ -0,0 +1,259 @@
...
+                                                className="btn btn-outline-primary"
+                                                value={subscribed ? gettext('Unsubscribe') : gettext('Subscribe')}
+                                                onClick={this.onChange}
+                                                disabled={this.state.saving} />
+                                        </div>
+                                    )}
+                                    <TextInput
+                                        name='subject'
+                                        label={gettext('Subject line')}
+                                        value={watchList.subject}
```
MarkLark86: If `watchList` is null then we should not be rendering the input elements until it has a value. It is causing the following warning in React:

`Warning: A component is changing an uncontrolled input of type text to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components`

assets/user-profile/components/UserProfileApp.jsx
```diff
@@ -22,16 +22,13 @@ import FollowedTopics from 'search/components/FollowedTopics';
 import UserProfileMenu from './UserProfileMenu';
 import UserProfileAvatar from './UserProfileAvatar';
 import ShareItemModal from 'components/ShareItemModal';
+// import UserWatchListModal from 'components/UserWatchListModal';
```
MarkLark86: Remove this comment

newsroom/watch_lists/views.py
```diff
@@ -71,22 +71,36 @@ def create():
...
 
     if flask.request.method == 'POST':
         form = WatchListsForm(watch_list=watch_list)
         if form.validate_on_submit():
-
             updates = form.data
+            request_updates = flask.request.get_json()
+
+            # If the updates have anything other than 'users', only admin or watch_list_admin can update
+            if len(request_updates.keys()) == 1 or 'users' not in request_updates:
```
MarkLark86: This seem a bit generic, i.e. len(keys) == 1 or 'users'. Would it be better to check equality with `keys == ['users']`?

PR #887, opened 2019-11-04
fix(API item expiry) stop the API returning expired items

marwoodandrew:


newsroom/news_api/formatters/service.py
```diff
@@ -33,7 +35,10 @@ def get_version(self, id, version, formatter_name):
             item['_id'] = item[id_field]
         else:
             item = get_resource_service('items').find_one(req=None, _id=id)
-        if not item:
+            if not item:
+                abort(404)
+        # Ensure that the item has not expired
+        if utcnow() - timedelta(days=app.config.get('MAX_EXPIRY_QUERY_LIMIT')) > item.get('versioncreated', utcnow()):
```
MarkLark86: There is a specific config for the News API `news_api_time_limit_days` (https://github.com/superdesk/newsroom/blob/master/newsroom/news_api/__init__.py#L10), and a utility method for retrieving this setting.

```python
from newsroom.settings import get_setting
get_setting('news_api_time_limit_days')
```

newsroom/news_api/app/__init__.py
```diff
@@ -43,7 +43,8 @@ def get_app(config=None):
               data=SuperdeskDataLayer,
               settings=app_config)
 
-    app._general_settings = {}
+    app._general_settings = {'news_api_time_limit_days': {'type': 'number',
```
MarkLark86: `news_api_time_limit_days` is defined in `news_api/__init__.py`. If you  add `newsroom.news_api` to the `CORE_APPS` in `settings.py` this setting will be registered in one place instead of two.

marwoodandrew: It's only defined in the parameters passed to a Newsroom app, I'm not sure how I can get to that!

PR #889, opened 2019-11-06
[SDAN-440] Include 'keywords' in watch_list creation and other bug-fixes around watch_list settings feature

nrvikas:


assets/components/TagList.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {get} from 'lodash';
+
+export default function TagList({tags, onClick, icon}) {
+    return ((get(tags, 'length', 0) > 0 && <div className='tag-list'>
+        <ul>
+            {tags.map((t, index) => (
+                <li key={index} onClick={onClick ? onClick.bind(null, index) : null}>
```
MarkLark86: Add the `noselect` class so the text doesn't get selected when removing a keyword

PR #905, opened 2019-11-26
[SDCP-110] - Adding optional preview configuration for actions in newshub

akintolga:


assets/layout/components/BaseApp.jsx
```diff
@@ -96,8 +96,9 @@ export default class BaseApp extends React.Component {
         }
     }
 
-    filterActions(item) {
-        return this.props.actions.filter((action) => !action.when || action.when(this.props.state, item));
+    filterActions(item, config) {
```
nrvikas: This can probably be a utils function as it's definition repeated many times.

akintolga: You're right Vikas, I'll merge for now but will create another PR for the refactoring.

PR #913, opened 2019-12-03
[SDAN-620]  Improvements to coverage inquiry email

nrvikas:


newsroom/templates/coverage_request_content.txt
```diff
@@ -0,0 +1,4 @@
+{% block content %}
+A coverage inquiry has been submitted by {{ name }} ({{ email }}) from {{ company }} regarding Agenda item: <a href="{{url}}">{{ url }}
```
MarkLark86: The link in this template should be plain text

newsroom/templates/coverage_request_content.html
```diff
@@ -0,0 +1,7 @@
+{% block content %}
+<p>A coverage inquiry has been submitted by {{ name }} ({{ email }}) from {{ company }} regarding Agenda item: <a href="{{url}}">{{ url }}</a></p>
+
+Message:
+{% include "message.html" %}
+<a href="mailto:{{ email }}?cc={{ recipients|join(',') }}&subject={{ subject }}">Click to reply</a>
```
MarkLark86: There should probably be some spacing between the message and the `Click to reply` link

nrvikas: There is a sufficient gap in the actual email because of the padding of message.

newsroom/templates/coverage_request_content.txt
nrvikas: Why should a URL link be plain text?

nrvikas: Yup, just realized it is a .txt. Will correct it. Thanks.

PR #920, opened 2019-12-05
fix tests setup

petrjasek:
- generate single bundle


MarkLark86:
@petrjasek It looks like the CI build is not quite fixed (maybe fireq again?)

When trying to log into any PR, we get the following error:
`csrf_token: The CSRF token is invalid.`

petrjasek:
@MarkLark86 I think https://github.com/superdesk/fireq/pull/56 will fix it

PR #925, opened 2019-12-09
Footer static pages

darconny:


newsroom/public/views.py
```diff
@@ -2,6 +2,12 @@
 from newsroom.public import blueprint
 
 
-@blueprint.route('/copyright')
-def copyright():
-    return flask.render_template('copyright.html')
+@blueprint.route('/privacy')
+def privacy():
+    return flask.render_template('privacy.html')
+@blueprint.route('/terms')
```
akintolga: I think you need 2 blank spaces before @blueprint otherwise flake8 will faill

theme/footer.html
```diff
@@ -3,9 +3,8 @@
         <img src="{{ url_for('theme', filename = 'login-logo-footer.svg') }}" class="d-none d-lg-block mr-3" />
     </div>
     <div>
-        {% if config.PRIVACY_POLICY %}<a href="{{ config.PRIVACY_POLICY }}" target="_blank">{{ gettext('Privacy Policy') }}</a>{% endif %}
-        {% if config.TERMS_AND_CONDITIONS %}<a href="{{ config.TERMS_AND_CONDITIONS }}" target="_blank">{{ gettext('Terms and Conditions') }}</a>{% endif %}
-        {% if config.CONTACT_ADDRESS %}<a href="{{ config.CONTACT_ADDRESS }}" target="_blank">{{ gettext('Contact Us') }}</a>{% endif %}
-        {% if config.SHOW_COPYRIGHT %}<a href="{{ url_for('public.copyright') }}" target="_blank">{{ gettext('Copyright') }}</a>{% endif %}
+        {% if config.SHOW_PRIVACY_POLICY %}<a href="{{ url_for('public.privacy') }}" target="_blank">{{ gettext('Privacy Policy') }}</a>{% endif %}
+        {% if config.SHOW_TERMS %}<a href="{{ url_for('public.terms') }}" target="_blank">{{ gettext('Terms and Conditions') }}</a>{% endif %}
+        {% if config.SHOW_CONTACT %}<a href="{{ url_for('public.contact') }}" target="_blank">{{ gettext('Contact Us') }}</a>{% endif %}
```
petrjasek: I guess you don't need those if's, that's for configuring core template, but this one is custom anyhow

darconny: Thanks, OK I will remove if's and leave only <a> tags.

newsroom/public/views.py
darconny: Thanks I will try adding 2 blank spaces before @blueprint.

newsroom/public/views.py
```diff
@@ -2,6 +2,12 @@
 from newsroom.public import blueprint
 
 
-@blueprint.route('/copyright')
-def copyright():
-    return flask.render_template('copyright.html')
+  @blueprint.route('/privacy')
```
petrjasek: this should be like:
```py
@blueprint.route('/privacy')
def privacy():
    return flask.render_template('privacy.html')


@blueprint.route('/terms')
def terms():
    return flask.render_template('terms.html')


@blueprint.route('/contact')
def contact():
    return flask.render_template('contact.html')
```
so 2 empty lines before each `@blueprint`

PR #931, opened 2019-12-12
Disabling watch lists for cp

akintolga:


settings.py
```diff
@@ -116,3 +112,11 @@
 }
 
 WATERMARK_IMAGE = None
+
+#: celery beat config
+CELERY_BEAT_SCHEDULE = {
```
petrjasek: imo it would be better to include it from default settings:
```py
from newsroom.default_settings import CELERY_BEAT_SCHEDULE as CELERY_BEAT_SCHEDULE_DEFAULT

CELERY_BEAT_SCHEDULE = {key: val for key, val in CELERY_BEAT_SCHEDULE_DEFAULT if key == 'newsroom.company...'}
```

newsroom/users/views.py
```diff
@@ -37,7 +36,6 @@ def get_view_data():
         'topics': get_user_topics(user['_id']) if user else [],
         'companyName': get_user_company_name(user),
         'locators': get_vocabulary('locators'),
-        'watch_lists': get_watch_lists_for_company(user),
```
petrjasek: can you make it somehow configurable? this will create conflicts imo

akintolga: I have added a parameter for now to get CP to prod, I am creating the same parameter change against master to keep it consistent

MarkLark86:
@akintolga @petrjasek The `News API` and `Watch Lists` are features that are still in development. As such, I recommend branching CP (and others) from the version branches which do not include these features yet. I will be creating the 1.17 branch today.

Any requirement to make these features optional should be placed in master, which will make your lives easier when it comes to maintenance and support.

akintolga:
I agree with you @MarkLark86 Still I need to merge this for now due to the deadline of this project. I checked branch 1.17, it was a little behind so can't rebase on that yet. I will create a PR against master to add the option over there in the next few weeks if that works for you.

PR #943, opened 2020-01-06
[SDAN-624] Display Take Key when Downloading, Copying, Printing and Sharing wire items

nrvikas:


newsroom/push.py
```diff
@@ -131,6 +131,7 @@ def publish_item(doc, is_new):
         if agenda_items:
             [notify_new_item(item, check_topics=False) for item in agenda_items]
     publish_item_signal.send(app._get_current_object(), item=doc, is_new=is_new)
+    doc['anpa_take_key'] = '4th update'
```
MarkLark86: I'm guessing this is no longer required!

PR #983, opened 2020-02-19
[SDCP-146] - remove category and subject filters

akintolga:


MarkLark86:
The PR https://github.com/superdesk/newsroom/pull/973 (Use same werkzeug version as eve 0.7.8) should fix your failing tests.

PR #1006, opened 2020-03-17
[SDCP-149] - Filtering story image for stories with images

akintolga:


akintolga:
Yes, actually it is time to do that. Since Christmas last year there has been changes that is pushed in haste to hit some deadlines and I have been putting it off to do a catch up from master. 

So let me start merging master into cp-lji and start moving this change and others to master.

PR #1011, opened 2020-03-19
[SD-149] - image downloads for stories

akintolga:


newsroom/wire/formatters/picture.py
```diff
@@ -0,0 +1,19 @@
...
+    FILE_EXTENSION = 'jpg'
+
+    def format_item(self, item, item_type='items'):
+        if item_type == 'agenda':
+            raise TypeError('Undefined format for agenda')
+
+        picture = get_picture(item)
+
+        if not picture:
+            raise TypeError('Undefined picture')
```
petrjasek: this should be rather `ValueError` imo

newsroom/wire/formatters/picture.py
```diff
@@ -0,0 +1,19 @@
+from newsroom.wire.formatters.base import BaseFormatter
+from newsroom.wire.utils import get_picture
+
+
+class PictureFormatter(BaseFormatter):
+
+    MIMETYPE = 'image/jpeg'
+    FILE_EXTENSION = 'jpg'
```
petrjasek: afaik they work with png as well, I think there is mimetype on the item, so we will have to make it dynamic probably

newsroom/wire/views.py
```diff
@@ -182,7 +183,14 @@ def download(_ids):
     formatter = app.download_formatters[_format]['formatter']
     mimetype = None
     attachment_filename = '%s-newsroom.zip' % utcnow().strftime('%Y%m%d%H%M')
-    if len(items) == 1 or _format == 'monitoring':
+    if _format == 'picture':
```
petrjasek: would be good to do this based on some formatter property

assets/components/SelectInput.jsx
```diff
@@ -26,7 +27,7 @@ const SelectInput = ({
                         <option value="">{defaultOption}</option>
                     }
                     {options.map((option) => {
-                        return <option key={option.value} value={option.value}>{option.text}</option>;
+                        return <option key={option.value} value={option.value}>{gettext(option.text)}</option>;
```
petrjasek: I think here the text should be translated already

newsroom/wire/views.py
```diff
@@ -182,7 +185,30 @@ def download(_ids):
...
+                    url_for('upload.get_upload',
+                            media_id=picture['media'],
+                            filename='baseimage%s' % formatter.get_file_extension()))
+            except ValueError:
+                return flask.abort(404)
+        else:
+            with zipfile.ZipFile(_file, mode='w') as zf:
+                for item in items:
+                    try:
+                        picture = formatter.format_item(item, item_type=item_type)
```
petrjasek: seems like this is not used, imo it should be rather:
```py
formatter.get_file_extension(item)
```

newsroom/upload.py
```diff
@@ -48,7 +48,7 @@ def get_upload(media_id):
 
     if flask.request.args.get('filename'):
         response.headers['Content-Type'] = media_file.content_type
-        response.headers['Content-Disposition'] = 'attachment; filename="%s"' % flask.request.args['filename']
+        response.headers['Content-Disposition'] = 'attachment; filename=%s' % flask.request.args['filename']
```
petrjasek: why removing these? [seems like it should be there](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition)

akintolga: I noticed that `flask.send_file` don't use quotations so that's the reason I removed here. But I'll put it back.

newsroom/wire/views.py
akintolga: Passing the file_extension with picture

assets/components/SelectInput.jsx
akintolga: Removed, but once I removed it I saw the dropdown became untranslated, will check once more after the merge

PR #1017, opened 2020-03-23
add simple search bar to dashboard

petrjasek:
SDCP-142

### DIFF
--- a/assets/home/components/search-bar.jsx
+++ b/assets/home/components/search-bar.jsx
@@ -4,23 +4,27 @@ import {gettext} from 'utils';
 export class SearchBar extends React.PureComponent {
     render() {
         return (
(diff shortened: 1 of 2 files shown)


assets/home/components/search-bar.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import {gettext} from 'utils';
+
+export class SearchBar extends React.PureComponent {
+    render() {
+        return (
+            <nav className="content-bar navbar justify-content-start flex-nowrap flex-sm-wrap">
```
darconny: Please put nav inside the block: 
`<section class="content-header"><nav ...>...</nav></section> `
so we have proper search bar height (56px).

darconny: That's it thanks. If it is not too much of work can you also add 'x' for clearing the search input, same as in wire view?

PR #1020, opened 2020-03-24
[SDCP-182] - Language in the url from emails

akintolga:


petrjasek:
I think we might need the same on the english links so we don't have the same issue there

PR #1024, opened 2020-03-25
Wire item image aspect ratio fix

darconny:
image aspect ratio in wire item, using preview image instead of thumbnail rendition

### DIFF
--- a/assets/styles/index.scss
+++ b/assets/styles/index.scss
@@ -1614,9 +1614,11 @@ article.list {
             }
             img {
                 max-width: none;
+                max-height: none;
                 width: 100%;
                 @include xl {
                     max-width: 160px;
+                    max-height: 140px;
                     width: auto;
                 }
             }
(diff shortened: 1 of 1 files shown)


assets/wire/components/WireListItem.jsx
```diff
@@ -4,7 +4,7 @@ import classNames from 'classnames';
 import {get} from 'lodash';
 
 import {gettext, fullDate, wordCount, LIST_ANIMATIONS, getSlugline} from 'utils';
-import {getPicture, getThumbnailRendition, showItemVersions, shortText, isKilled, getVideos} from 'wire/utils';
+import {getPicture, getPreviewRendition, showItemVersions, shortText, isKilled, getVideos} from 'wire/utils';
```
petrjasek: why do you need this change? I don't think we should do this, thumbnail rendition is configured not to include watermark, while preview could have it.

darconny: Because thumbnail rendition is cutting images to 640x480 and we need original aspect ratio. In that case please make a logic for generating this kind of rendition, which maintains original image aspect ratio.

petrjasek: that was just temporary, new images should have original aspect ration for thumbnails

darconny: Ok thanks for clarifying, I left only css part.

PR #1025, opened 2020-03-26
[SDCP-201] - Removing image if the update doesn't have it

akintolga:


newsroom/push.py
```diff
@@ -132,6 +133,8 @@ def publish_item(doc, is_new):
             [notify_new_item(item, check_topics=False) for item in agenda_items]
     publish_item_signal.send(app._get_current_object(), item=doc, is_new=is_new)
     _id = service.create([doc])[0]
+    if 'associations' not in doc and not is_new and previous_associations_exists:
+        service.patch(_id, updates={'associations': {'featuremedia': None}})
```
petrjasek: that won't work for other media like pictures in embedded in body, what about passing orig file into publish item and check old associations:
```py
def publish_item(doc, orig=None):
    ...

    if orig.get('associations'):
        for assoc in orig['associations']:
            updates['associations'][assoc] = None
    ...
```

akintolga: What if I do this, do you think would it work?
`service.patch(_id, updates={'associations': None})`

petrjasek: that should work too

newsroom/push.py
```diff
@@ -101,7 +102,7 @@ def set_dates(doc):
     doc.setdefault(app.config['VERSION'], 1)
 
 
-def publish_item(doc, is_new):
+def publish_item(doc, is_new, previous_associations_exists):
```
petrjasek: I would rather have single `orig` param instead of those 2

PR #1033, opened 2020-04-07
Allow date format override depending on language

pablopunk:
SDCP-197
SDCP-207


newsroom/default_settings.py
```diff
@@ -255,21 +250,36 @@
...
+        "COVERAGE_DATE_FORMAT": "DD/MM",
+    },
+    "fr_CA": {  # example - you can overwrite any format above
+        "DATE_FORMAT": "DD/MM/YYYY",
+    }
+}
+CLIENT_TIME_FORMAT = 'HH:mm'
+CLIENT_DATE_FORMAT = 'DD/MM/YYYY'
+CLIENT_COVERAGE_DATE_TIME_FORMAT = 'HH:mm DD/MM'
+CLIENT_COVERAGE_DATE_FORMAT = 'DD/MM'
```
petrjasek, line 270: not sure if we still need these, is it used anywhere?

pablopunk: No we don't, I guess I didn't stage that change, I'll fix that 👍

newsroom/default_settings.py
```diff
@@ -255,21 +250,32 @@
...
+    'locale_formats': CLIENT_LOCALE_FORMATS,
     'coverage_types': COVERAGE_TYPES,
     'display_abstract': DISPLAY_ABSTRACT,
     'list_animations': True,  # Enables or disables the animations for list item select boxes,
     'display_news_only': True,  # Displays news only switch in wire,
     'default_timezone': DEFAULT_TIMEZONE
 }
 
-LANGUAGES = ['en', 'fi', 'cs', 'fr_CA']
-DEFAULT_LANGUAGE = 'en'
```
petrjasek, line 272: why removing these?

pablopunk: I just moved them [below](https://github.com/superdesk/newsroom/pull/1033/files#diff-a9726395d27997f1b1342236c7e5539fR253-R254), so it's near the client config

PR #1035, opened 2020-04-08
[STTNHUB-65] - Filter embargoed items from cards

akintolga:


petrjasek:
so what's the idea? display embargoed items by default?

akintolga:
I think idea is to use company type filter for home page cards as well. At the moment it is only used in wire and it is working.

newsroom/wire/search.py
```diff
@@ -282,22 +286,17 @@ def has_permissions(self, item, ignore_latest=False):
...
         if not product:
             return
 
-        query['bool']['must'].append({
-            "bool": {
-                "should": [
-                    {"range": {"embargoed": {"lt": "now"}}},
-                    {"bool": {"must_not": {"exists": {"field": "embargoed"}}}}
-                ]
-            }
```
petrjasek, line 299: so how is this filter applied now?

akintolga: There's a filter in company_types https://github.com/superdesk/newsroom/blob/stt/settings.py#L61
So this one will be used now. But let me know if you think we need another way.

petrjasek: I see, np then

PR #1036, opened 2020-04-08
Format dates with current locale

pablopunk:
SDCP-207

moment.js already provides abbreviations for months

### DIFF
--- a/assets/utils.js
+++ b/assets/utils.js
@@ -727,5 +727,5 @@ export function getSlugline(item, withTakeKey = false) {
  */
 export function isActionEnabled(configKey) {
     const config = getConfig(configKey, {});
-    return action => config[action.id] == null || config[action.id];
+    return (action) => config[action.id] == null || config[action.id];
 }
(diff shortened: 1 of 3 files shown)


assets/utils.js
```diff
@@ -17,6 +18,10 @@ import {
     TO_BE_CONFIRMED_TEXT
 } from './agenda/utils';
 
+// Add locales to moment.js
+moment.locale('fr_CA');
+moment.locale('en');
```
petrjasek, line 23: shouldn't it be based on the config rather than hardcoding it here?

pablopunk: I tried to load the config from `window` object here but it's not available yet

petrjasek: ok

pablopunk: also if we add another locale, we would still need to add it manually [here](https://github.com/superdesk/newsroom/pull/1036/files#diff-c6a35b21c13bfe61f8ac39d33b8d771dR12) cause that's how moment works 🤷‍♂

petrjasek: would be good to leave a comment regarding this imo

PR #1040, opened 2020-04-24
fix featuremedia picture not visible for story

petrjasek:
in case there were some pictures in the body first one would be used instead of featuremedia

SDCP-214

### DIFF
--- a/assets/wire/utils.js
+++ b/assets/wire/utils.js
@@ -41,7 +41,9 @@ export function getVideos(item) {
  * @return {Object}
  */
 export function getPicture(item) {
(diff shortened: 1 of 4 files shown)


MarkLark86:
@petrjasek I have concerns with the fixes that have been going into cp-lji that are not being cherry-picked into master. These two branches are starting to diverge and it is going to get harder and harder to maintain.

Fixes should go into master, then cherry-picked into the respective customer branches to make sure master includes all changes (non customer specific)

petrjasek:
yep right, will do that

PR #1045, opened 2020-04-29
fix: Missing userType to add remove action

pablopunk:
... to home items preview

SDESK-5201


MarkLark86:
The remove action is not visible in the home page in the test instance. `userType` also needs to be populated from the server for the home view and then stored in the Redux store on `INIT_DATA` for the home page.

pablopunk:
not sure why tests are failing but it's also happening on master

PR #1054, opened 2020-05-29
[SDAN-275] Highlighting wire searches

marwoodandrew:


newsroom/search.py
```diff
@@ -303,6 +308,26 @@ def prefill_search_items(self, search):
                 {'constant_score': {'filter': {'exists': {'field': 'nextversion'}}}}
             )
 
+    def prefill_search_highlights(self, search, req=None):
+        if req is None:
+            return
+        query_string = req.args.get('q')
```
MarkLark86, line 314: `args` are already placed on the search instance under `search.args` (https://github.com/superdesk/newsroom/blob/ee3d799953534a47426182c2005574b92fcf5cb1/newsroom/search.py#L182)

PR #1055, opened 2020-06-05
[SDAN-642] Embed logo into RTF monitoring reports

marwoodandrew:


marwoodandrew:
I could move this out of the newsroom repo and implement it in the "customize_rtf_file" function?

PR #1059, opened 2020-06-23
Merging master into STT

akintolga:


petrjasek:
would be good to fix those tests, also we shouldn't squash this so next time it won't apply those commits again, so pls merge this manually

PR #1062, opened 2020-07-30
Display credits info on picture view/preview

ride90:
SDESK-5310
SDAN-613 (just to link an old ticket to this PR)

- use redux-dev tools for non-production instances
- use `display_credits` (default to false) config to display credits under the byline on preview/detail


ride90:
@marwoodandrew will it be an issue for AAP if I leave a byline as it is now and add copyrightholder + copyrightnotice on the new line?

PR #1074, opened 2020-09-04
ES7 support, use 2.0 cores

ride90:
IN-4355

fireq pr https://github.com/superdesk/fireq/pull/78


petrjasek (approved):

newsroom/push.py
```diff
@@ -109,6 +109,8 @@ def publish_item(doc, original):
     doc.setdefault('wordcount', get_word_count(doc.get('body_html', '')))
     doc.setdefault('charcount', get_char_count(doc.get('body_html', '')))
     service = superdesk.get_resource_service('content_api')
+    service.datasource = 'items'
```
MarkLark86, line 112: Why is the `datasource` being set here? Would it be better to perform this change on an `init_app` somewhere?

ride90: Hi Mark, 

When I ingest item from SD to NH, item appears in DBs (mongo and elastic), in elastic that item has `"_resource": "content_api"` and it does not appear in the wire, but if I rebuild index, that item appears in the wire search and in elastic it has a proper `"_resource": "items"`.

The reason is that when we register a service, we provide `endpoint_name = 'content_api'` and it’s used as datasource which is wrong, basically endpoint name and datasource are different things.
This fix is temporary and here is a ticket to provide a core fix https://dev.sourcefabric.org/browse/SDESK-5529

Answering your question, I think init_app here is a part of SD core.

ride90: @MarkLark86 merging it for now, so QA can test manually.

PR #1076, opened 2020-10-02
News API add ability to retrieve images

marwoodandrew:


marwoodandrew:
@MarkLark86 , @ride90 Any one want to take a look at this? If you're happy I'll cherry pick it into develop!

newsroom/news_api/utils.py
```diff
@@ -33,3 +33,17 @@ def format_report_results(search_result, unique_endpoints, companies):
...
     return results
+
+
+def remove_internal_renditions(item):
+    clean_renditions = dict()
+
+    # associations featuremedia will contain the internal newsroom renditions, we need to remove these.
+    if ((item.get('associations') or {}).get('featuremedia') or {}).get('renditions'):
+        for key, rendition in\
+                (((item.get('associations') or {}).get('featuremedia') or {}).get('renditions')).items():
```
ride90, line 44: here we can make it more readable:
```
item['associations']['featuremedia']['renditions'].items()
```

ride90 (approved):

newsroom/news_api/news/assets/assets.py
```diff
@@ -0,0 +1,38 @@
...
+from newsroom.upload import ASSETS_RESOURCE
+from flask_babel import gettext
+import bson.errors
+from werkzeug.wsgi import wrap_file
+from newsroom.news_api.utils import post_api_audit
+
+blueprint = superdesk.Blueprint('assets', __name__)
+
+
+@blueprint.route('/{}/assets/<path:asset_id>'.format(URL_PREFIX), methods=['GET'])
```
MarkLark86, line 15: Instead of setting `URL_PREFIX` here it might be better to add this in the factory method that registers the blueprints: https://github.com/superdesk/newsroom/blob/master/newsroom/factory/app.py#L132

Similar to how it is done in Superdesk (by getting `api_prefix` from the app at runtime):
https://github.com/superdesk/superdesk-core/blob/develop/superdesk/__init__.py#L90

This would also have to be changed for `newsroom.news_api.news.item.item`
https://github.com/superdesk/newsroom/blob/master/newsroom/news_api/news/item/item.py#L13

MarkLark86 (approved):
Looks good, just one improvement suggestion for registering blueprints and the `URL_PREFIX` system setting

newsroom/news_api/utils.py
marwoodandrew: Done

newsroom/news_api/news/assets/assets.py
marwoodandrew: Done

marwoodandrew:
@MarkLark86 I believe I've made the changes you suggested, would appreciate a quick check I've done what you suggested!

MarkLark86 (approved):

PR #1084, opened 2020-10-28
fix(API) stop returning products, media and subscribers with image me…

marwoodandrew:
…ta data

### DIFF
--- a/features/news_api_item.feature
+++ b/features/news_api_item.feature
@@ -113,9 +113,12 @@ Feature: News API Item
       "body_html": "<p>test&nbsp;test</p>",
       "associations": {
         "featuremedia": {
+          "products": ["12345", "6789"],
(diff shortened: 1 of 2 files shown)


features/news_api_item.feature
```diff
@@ -113,9 +113,12 @@ Feature: News API Item
       "body_html": "<p>test&nbsp;test</p>",
       "associations": {
         "featuremedia": {
+          "products": ["12345", "6789"],
```
MarkLark86, line 116: You could also add to the response of `GET "/news/item/#items._id#?format=NINJSFormatter2"`
```
"products": "__no_value__",
"subscribers": "__no_value__"
```
To make sure these keys don't exist in the response

ride90 (approved):

PR #1090, opened 2020-11-09
Github actions

ride90:
- github actions
- fix linter/pytest errors
- fix reportlab version to have support for python 3.5

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -62,7 +62,8 @@ jobs:
             ${{ runner.os }}-npm
             ${{ runner.os }}-
 
-      - run: npm install --no-audit
+      - name: npm ci
+        run: npm ci
 
       - name: lint
         run: npm run lint
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,74 @@
+name: "CI"
+
+on:
+  [push, pull_request]
+
+jobs:
+
+  server:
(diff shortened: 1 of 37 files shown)


petrjasek (approved):

.github/workflows/tests.yml
```diff
@@ -0,0 +1,74 @@
...
+      - name: cache npm
+        uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-npm-${{ hashFiles('package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-npm
+            ${{ runner.os }}-
+
+      - run: npm install --no-audit
```
petrjasek, line 65: if you have package-lock you can use `npm ci` should be a bit faster

ride90: ohh right

requirements.txt
```diff
@@ -1,6 +1,8 @@
 httmock
 wooper
 PyRTF3
+# reportlab is fixed due to support for python 3.5
+reportlab==3.5.53
```
MarkLark86, line 5: This makes sense in the `master` branch, when this is ported to the `develop` we probably should bump Python to 3.6+

MarkLark86 (approved):

requirements.txt
marwoodandrew: This one caught us out on our last release!

PR #1094, opened 2020-11-11
Fix urllib3 version

ride90:


requirements.txt
```diff
@@ -3,6 +3,7 @@ wooper
 PyRTF3
 xhtml2pdf
 werkzeug>=0.9.4,<=0.11.15
+urllib3<1.26
```
MarkLark86, line 6: It would be good to add this to `setup.py:install_requires` as well.

ride90: done

MarkLark86 (approved):

PR #1096, opened 2020-11-12
Port initialize_data from superdesk-core

ride90:


newsroom/initialize_data.py
```diff
@@ -0,0 +1,52 @@
...
+        # create indexes in mongo
+        app.init_indexes()
+        # put mapping to elastic
+        app.data.init_elastic(app)
+
+        if init_index_only:
+            logger.info('Only indexes initialized.')
+            return 0
+
+        print(DATA_PATH)
```
petrjasek, line 32: ups

petrjasek (approved):

PR #1110, opened 2020-11-25
feat(wire): Improve metadata UI

pablopunk:
STTNHUB-105
STTNHUB-106
STTNHUB-112
STTNHUB-113
STTNHUB-114
STTNHUB-115
STTNHUB-116

All tasks include improvements for the metadata UI, this PR handles them all


petrjasek:
I don't like those custom stt fields being hardcoded, can we introduce some api for that?

pablopunk:
@petrjasek should I make a new entry in ui_config with the fields just like we do on superdesk.config? Then I could match each field to react component

ride90:
Merging this one as it is for now. @pablopunk you can make it work via ui_config later.

PR #1112, opened 2020-12-01
Change format for displaying token expiry time in email notifications

ride90:
STTNHUB-121

Not sure if my translation is correct :) 

⚠️ Dependencies updates:
- use `superdesk-core==1.33.9`
- `superdesk-planning 1.10` -> `superdesk-planning 1.33.2` with icalendar4
- remove duplicated dependencies

⚠️ Tests updates:
- stateless tests. explicitly remove DB/Index
- `core` folder for core tests to avoid implicit usage of autofixtures in subfolders like `aap`


pablopunk (approved):

MarkLark86:
Planning 1.33.2 has now been released to pypi (with the upgraded `icalendar`).
So in `setup.py` and `requirements.txt` you can use `'superdesk-planning==1.33.2'`

requirements.txt
```diff
@@ -1,11 +1,3 @@
...
-wooper
-PyRTF3
-reportlab==3.5.53
-xhtml2pdf
-werkzeug>=0.9.4,<=0.11.15
-email-validator==1.0.5
-urllib3<1.26
 -e .
-git+git://github.com/superdesk/superdesk-planning.git@1.10#egg=superdesk-planning
+superdesk-planning==1.33.2
```
MarkLark86, line 2: It might be better to have these two dependencies in `setup.py:install_requires`, especially if we publish this to PyPi as a package (example usage would be for a customer repo)

ride90: done

petrjasek (approved):

PR #1116, opened 2020-12-09
Fix previous version view.

ride90:
related to STTNHUB-40


assets/wire/components/WireListItem.jsx
```diff
@@ -315,6 +315,7 @@ class WireListItem extends React.Component {
                     <ListItemPreviousVersions
                         item={this.props.item}
                         isPreview={false}
+                        displayConfig={this.props.listConfig}
```
pablopunk, line 318: shouldn't this be `this.props.displayConfig`?

ride90: no, it should be a `listConfig`

pablopunk (approved):

PR #1118, opened 2020-12-11
feat(wire): Custom metadata fields

pablopunk:
> I changed some eslint rules to make the code more consistent, so tests are failing now. When the PR is reviewed I'll fix the lint so the `git diff` is not so big for reviews

### Example config

Ilustración_sin_título


pablopunk:
The separators (`//`) are still hardcoded, I'm thinkging about allowing configs like: `["source", "//", "charcount", "/", "wordcount"]`

petrjasek:
I wouldn't mind making it more verbose with just 1 way of doing things, we don't change this often, something like:
```ts
fields: [
    [{"field": "urgency"}],
    [{"field": "source"}, {"field": "sttdepartment", ...}],
    [{"field": "charcount"}, {"field": "wordcount"}],
],
```

pablopunk:
@petrjasek isn't it a bit weird to have all of them be arrays? Also, it wouldn't make the code much simpler, it's not too complicated right now

petrjasek:
yep if it's not about having lines but rather this whole thing is a sequence of fields it is better without arrays

pablopunk:
@petrjasek what about the separators (`//`), should I keep them hardcoded as they are right now?

assets/wire/components/fields/CharCount.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import {characterCount, gettext} from 'utils'
```
petrjasek, line 3: would be good to keep it consistent and have `;` here, probably with eslint rule to check it

assets/wire/components/fields/index.jsx
```diff
@@ -0,0 +1,164 @@
...
+            Component = WordCount;
+            break;
+        case 'previous_versions':
+            Component = PreviousVersions;
+            break;
+        case 'embargo':
+            Component = Embargo;
+            break;
+        case 'versioncreated':
+            Component = VersionCreated;
```
petrjasek, line 142: imo would be good to have some mapping for components so it would be easy to add a new type or override the core implementation with something custom

assets/wire/components/fields/CharCount.jsx
pablopunk: the rule already exists on our config, I'll run `eslint --fix` once the PR is reviewed 👍🏻

assets/wire/components/fields/index.jsx
pablopunk: like an object or something?

```ts
{
  versioncreated: VersionCreated,
  ...
}
```

petrjasek: exactly

petrjasek (approved):

PR #1122, opened 2021-01-07
Disable selection config for wire preview 

pablopunk:
SDAN-694


setup.py
```diff
@@ -2,23 +2,20 @@
 
 install_requires = [
     'Babel>=2.5.3,<3.0',
-    'eve==0.7.8',
-    'eve-elastic==2.5.0',
-    'WTForms==2.2.1',
```
ride90, line 7: I think we need this one.

pablopunk: ah you're right

PR #1130, opened 2021-03-30
fix management pages, stop returning bson

marwoodandrew:
Depends on https://github.com/superdesk/superdesk-core/pull/2182

### DIFF
--- a/newsroom/template_filters.py
+++ b/newsroom/template_filters.py
@@ -4,7 +4,6 @@ import flask
 import hashlib
 
 from flask import current_app as app
-from bson import json_util as bson_json_util
 from eve.utils import str_to_date
 from flask_babel import format_time, format_date, format_datetime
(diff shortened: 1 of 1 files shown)


newsroom/template_filters.py
```diff
@@ -15,8 +15,7 @@
 
 def to_json(value):
     """Jinja filter to address the encoding of special values to json"""
-
-    return bson_json_util.dumps(value)
+    return SuperdeskJSONEncoder().dumps(value)
```
MarkLark86, line 18: It might be better to use `app.json_encoder().dumps(value)` so that this code will use whatever is registered when bootstrapping the app (in case the encoder changes, for whatever reason).

marwoodandrew: Done!

MarkLark86 (approved):

PR #1134, opened 2021-04-22
[SDAN-702] Allow account managers to do some administrative task

marwoodandrew:
Working on some tests!

### DIFF
--- a/assets/users/components/EditUser.jsx
+++ b/assets/users/components/EditUser.jsx
@@ -6,7 +6,7 @@ import CheckboxInput from 'components/CheckboxInput';
 import AuditInformation from 'components/AuditInformation';
 
 import { gettext } from 'utils';
-import {userTypes} from '../utils';
+import {userTypes, isUserAdmin} from '../utils';
 
 const getCompanyOptions = (companies) => companies.map(company => ({value: company._id, text: company.name}));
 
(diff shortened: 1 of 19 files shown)


marwoodandrew:
Ping @petrjasek @MarkLark86 Since we now have fewer admin staff we're trying to offload some of the admin to the account managers.

newsroom/templates/settings.html
```diff
@@ -16,11 +16,13 @@
         <div class='side-navigation' id='settings-menu'>
             <ul>
                 {% for app in settings_apps|sort(attribute='weight') %}
-                <li>
-                    <a class="side-navigation__btn {{ "active" if setting_type == app._id }}"
-                        href="{{ url_for('settings.app', app_id=app._id) }}">
-                        {{ gettext(app.name) }}</a>
-                </li>
+                    {% if is_admin() or ( is_admin_or_account_manager() and app.allow_account_mgr ) %}
```
petrjasek, line 19: would probably simplify the check in the template and keep the logic in python, maybe something like:
```
{% if is_authorized(app._id) %}
```
but maybe we can actually do the filtering on the settings_apps before going to template

assets/users/components/EditUser.jsx
```diff
@@ -69,15 +69,15 @@ function EditUser({user, onChange, errors, companies, onSave, onResetPassword, o
                         value={user.role}
                         onChange={onChange}
                         error={errors ? errors.role : null} />
-
+                    { currentUser.user_type === 'administrator' &&
```
MarkLark86, line 72: It would be good to have a utility function to determine if a user is an administrator, rather than checking against a string

newsroom/templates/settings.html
marwoodandrew: OK, added a function to return the authorized list of apps, that simplifies things.

assets/users/components/EditUser.jsx
marwoodandrew: Added a function in utils

MarkLark86 (approved):

PR #1139, opened 2021-06-17
SDAN-703 Apply product permissions to image assets available via the …

marwoodandrew:
…Newsroom API

### DIFF
--- a/newsroom/news_api/utils.py
+++ b/newsroom/news_api/utils.py
@@ -1,6 +1,7 @@
 from superdesk import get_resource_service
 from superdesk.utc import utcnow
 from flask import request, g
+from newsroom.products.products import get_products_by_company
 
 
 def post_api_audit(doc):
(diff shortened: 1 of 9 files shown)


marwoodandrew:
A simpler solution to this would have been to include the credit/source in the associations schema, but the schema is shared with superdesk and it would have been a core change specifically for AAP's use case. This solution moves the permissioning  into Superdesk.

marwoodandrew:
ping @petrjasek @MarkLark86

newsroom/news_api/utils.py
```diff
@@ -47,3 +48,25 @@ def remove_internal_renditions(item):
...
+    :return:
+    """
+    if ((item.get('associations') or {}).get('featuremedia') or {}).get('products'):
+        # Extract the products that the image matched in Superdesk
+        im_products = [p.get('code') for p in
+                       ((item.get('associations') or {}).get('featuremedia') or {}).get('products')]
+
+        # Check if the one of the companies products that has a superdesk product id matches one of the
+        # image product id's
+        sd_products = [p.get('sd_product_id') for p in get_products_by_company(g.user, None, 'news_api') if
```
MarkLark86, line 67: Does this assume that `sd_product_id` is used by the newsroom?
What happens if `sd_product_id` is not used in any products?

marwoodandrew: In the initial development there was a facility to allow Superdesk products to be reflected in Newsroom products by cutting and pasting the Superdesk product id into a Newsroom product, we never used it until now. The potential issue is that in Superdesk we generally don't restrict products to item types, so if we use this facility for text items we'll need to ensure all text products don't ever natch an image.

MarkLark86: I would say that would be an issue, forcing the use of product id's (for customers that don't want it).

This might be a good place for a system setting to turn this filter/restriction on/off

marwoodandrew: Fair point! I'll add a setting to disable/enable the checking!

MarkLark86 (approved):

PR #1140, opened 2021-06-23
fix user activity report when agenda is not enabled

petrjasek:
SDESK-6035

### DIFF
--- a/newsroom/reports/reports.py
+++ b/newsroom/reports/reports.py
@@ -204,10 +204,7 @@ def get_subscriber_activity_report():
 
         company_ids.append(ObjectId(doc.get('company')))
         user_ids.append(ObjectId(doc.get('user')))
-    try:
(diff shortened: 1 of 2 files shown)


MarkLark86:
@petrjasek It might help to fix this issue at the source (so this problem doesn't pop up elsewhere).

- https://github.com/superdesk/newsroom/blob/master/newsroom/utils.py#L80 - raising `abort(404)` on keyError
- https://github.com/superdesk/newsroom/blob/master/newsroom/utils.py#L310 - returning `None` on KeyError
- https://github.com/superdesk/newsroom/blob/master/newsroom/utils.py#L347 - returning `[]` on KeyError

It would also be benficial if this were merged into `master` / `develop` first, and cherry-picked into the `stt` branch

petrjasek:
yep you're right, I've updated the fix. will do the cherry-picking the other way around

marwoodandrew (approved):

MarkLark86 (approved):

PR #1146, opened 2021-08-31
fix picture download not working on chrome

petrjasek:
there was a redirect in a popup which got blocked
by recent chrome. now there is no popup which should
help in general and there is no redirect for media neither
just to be sure it won't be a problem in future.

SDCP-567


MarkLark86 (approved):
We should also look at fixing this in both `master` and `develop` branches, and for all download options to make sure this won't affect other instances/actions.

We should also look at fixing this in the `newsroom-core` repo, as we will be using this for CPs new instance

petrjasek:
yep we will push it everywhere eventually, but seems like only on cp there is this option to download image which is problematic now, other downloads should be working fine

PR #1150, opened 2021-11-15
Add admin settings to restrict the renditions the NewsAPI returns

marwoodandrew:


marwoodandrew:
@MarkLark86 Had to bump the timedelta in a test, passed for me but maybe in a slower environment would fail!

MarkLark86 (approved):

PR #1153, opened 2022-01-04
SDAN-710 Send monitoring alerts with articles in body

marwoodandrew:


marwoodandrew:
Ping! @MarkLark86 @petrjasek I showed the emails to Steve, he's OK'd it! There is an associated PR in newsroom-aap with the template, same template does the PDF and print.

MarkLark86 (approved):

PR #1158, opened 2022-02-21
SDAN-712 Monitoring emails are sent to users that are disabled

marwoodandrew:


newsroom/monitoring/email_alerts.py
```diff
@@ -239,6 +239,16 @@ def already_sent(self, item, profile):
 
         return get_resource_service('history').get(req=None, lookup=lookup).count()
 
+    def validate_users(self, m, company):
```
MarkLark86, line 242: The name is a bit misleading, as it's actually filtering and not validating the supplied args

MarkLark86 (approved):

PR #1166, opened 2022-09-30
Show embargo badge in item list

marwoodandrew:


assets/wire/components/WireListItem.jsx
```diff
@@ -100,6 +100,9 @@ class WireListItem extends React.Component {
                             {!isExtended && (
                                 <WireListItemIcons item={item} picture={picture} videos={videos} divider={false} />
                             )}
+                            {!isExtended && (
```
MarkLark86: Rather than duplicating the condition, you can use `React.Fragment`:
```typescript
{!isExtended && (
    <React.Fragment>
        <div>Render item 1</div>
        <div>Render item 2</div>
    </React.Fragment>
)}
```

marwoodandrew: @MarkLark86 I'd tried returning an array and that busted the client tests, I'll give the fragment a go!

MarkLark86 (approved):

PR #1169, opened 2023-07-05
Create the Lockhart feature

marwoodandrew:
Initial PR to impliment the Lockart feature, permissioning of embedded video, audio and images.

### DIFF
--- a/newsroom/templates/base_layout.html
+++ b/newsroom/templates/base_layout.html
@@ -123,16 +123,25 @@
   function gtag(){dataLayer.push(arguments);}
   gtag('js', new Date());
 
-  gtag('config', '{{ get_setting('google_analytics') | safe }}', {
-      custom_map: {
-          dimension1: 'company',
-          dimension2: 'user',
-      }
+  gtag('set', 'user_properties', {
+      company: (window.profileData || {}).companyName || 'none',
+      user: ((window.profileData || {}).user || {}).first_name || 'unknown',
   });
 
   {% if session.get('user') %}
   gtag('set', {'user_id': '{{ hash(session['user']) | safe }}' });
   {% endif %}
+
+  if ('{{ get_setting('google_analytics') }}'.startsWith('UA')) {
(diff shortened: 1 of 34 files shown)


newsroom/wire/formatters/utils.py
```diff
@@ -0,0 +1,140 @@
...
+                              ((item.get('associations') or {}).get(key) or {}).get('products', [])]
+
+            if not len(set(embed_products) & set(permitted_products)):
+                kill_keys.append(key)
+
+    # Nothing to do
+    if len(kill_keys) == 0:
+        return
+
+    root_elem = lxml_html.fromstring(item.get('body_html', ''))
```
MarkLark86, line 118: The code to iterate over embeds is used multiple times, might be good to have a function that allows to provide a callback for each group found

newsroom/wire/search.py
```diff
@@ -518,3 +525,59 @@ def get_matching_bookmarks(self, item_ids, active_users, active_companies):
                     bookmark_users.append(bookmark)
 
         return bookmark_users
+
+    def permission_embeds_in_item(self, item, permitted_products):
```
MarkLark86, line 529: Could this be purely done in the browser instead? I'm wondering about the performance cost here, and what are the pros/cons of doing this server side or browser side

MarkLark86 (approved):

newsroom/templates/base_layout.html
```diff
@@ -158,6 +167,11 @@
 <script src="//cdn.iframe.ly/embed.js" async></script>
 {% endif %}
 
+{% if config.PLYR %}
+<script src="//cdn.plyr.io/3.6.3/plyr.js"></script>
```
MarkLark86, line 171: I have reservations about using this library. It would be best if we could achieve this in pure HTML 5 players (or a player that is not deprecated), but for now it's okay

marwoodandrew: I'm hoping to get some time to do this, I think it will involve disabling the controls and implimenting buttons for the play ffwd, rew, download etc.

PR #1170, opened 2023-07-17
fix remove stray style on body

marwoodandrew:


assets/ui/components/ArticleBodyHtml.jsx
```diff
@@ -228,7 +228,7 @@ class ArticleBodyHtml extends React.PureComponent {
         return (
             <div
                 ref={this.bodyRef}
-                className='wire-column__preview__text wire-column__preview__text--pre'
+                className='wire-column__preview__text'
```
MarkLark86, line 231: @marwoodandrew What's the issue here? We're still using this with that latest version https://github.com/superdesk/newsroom-core/blob/develop/assets/ui/components/ArticleBodyHtml.jsx#L202

marwoodandrew: The but we're still using the old style definitions, it's messing with the paragraph styling!

MarkLark86 (approved):

PR #1171, opened 2023-07-17
fix logging download endpoints route getting confused by AWS storage folders

marwoodandrew:
The '/' in the media path confused the route so wouldn't match the endpoint

### DIFF
--- a/newsroom/history.py
+++ b/newsroom/history.py
@@ -7,11 +7,12 @@ from bson import ObjectId
 from superdesk import get_resource_service
 from superdesk.resource import not_analyzed, not_enabled
 from superdesk.utc import utcnow
-from flask import json, abort, Blueprint, jsonify, g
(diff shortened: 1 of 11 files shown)


newsroom/history.py
```diff
@@ -111,7 +112,48 @@ def create_media_history_record(self, item, association_name, action, user, sect
...
-    def log_media_download(self, item_id, media_id):
+    def _find_association(self, item, media_id):
+        """
+        Find the matching media association in the item
+        :param item: item object
+        :param media_id: ID of the media
+        :return: tuple (name, association) or a 404
+        """
+        for name, association in (item.get('associations') or {}).items():
+            for rendition in association.get("renditions"):
```
MarkLark86: It's plausible that `renditions` here could not be defined, in which you would get the following exception raised:

```bash
TypeError: 'NoneType' object is not iterable
```

MarkLark86 (approved):

PR #1180, opened 2023-10-13
fix cursor traversal when indexing

marwoodandrew:


MarkLark86 (approved):

MarkLark86:
@petrjasek We should probably port this one to `newsroom-core`. Not only does it add value to re-index from a particular ID (without destroying index), but it also has a performance improvement by not using `cursor.count()` which adds extra mongo queries to each iteration

PR #1181, opened 2023-10-16
Remove embeds from selected downloads

marwoodandrew:


newsroom/wire/formatters/ninjs2.py
```diff
@@ -18,3 +19,15 @@ def _transform_to_ninjs(self, item):
...
+
+class NINJSFormatter3(NINJSFormatter2):
+    """
+    Format with no Embeds
+    """
+
+    def _transform_to_ninjs(self, item):
+        remove_all_embeds(item)
+        ninjs = super()._transform_to_ninjs(item)
+        # do stuff
```
MarkLark86: Is there something missing here?

marwoodandrew: Maybe a return!

MarkLark86 (approved):

PR #1185, opened 2024-02-02
SDAN-725 homepage/notifications crash fix

marwoodandrew:


marwoodandrew:
I'm not in love with this, but sick of it crashing, so hit it with a big hammer!

PR #1191, opened 2024-08-06
Fix broken test build

marwoodandrew:


MarkLark86:
@marwoodandrew Wondering if this could be an issue with the version of pytest being used: https://github.com/superdesk/newsroom/blob/lockhart/dev-requirements.txt#L6. It is rather an old version

marwoodandrew:
@MarkLark86 Well all I seem to have done is find a sophisticated way to stop the tests from running, I'll look again next week. (still don't think the actual build will build due to the pip problem, but can't  see how I can fix it, I think It's building under python 3.10!

MarkLark86:
@marwoodandrew Looks like you're getting there. The `ipaddress` module doesn't seem to like `123.123.123.001` address.

```
>>> import ipaddress
>>> ipaddress.ip_network("123.123.123.001", strict=True)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/marklark/miniconda3/envs/async-nhub-tests/lib/python3.10/ipaddress.py", line 83, in ip_network
    raise ValueError(f'{address!r} does not appear to be an IPv4 or IPv6 network')
ValueError: '123.123.123.001' does not appear to be an IPv4 or IPv6 network
>>>
>>>
>>> ipaddress.ip_network("123.123.123.1", strict=True)
IPv4Network('123.123.123.1/32')
>>>
```

Try replacing the `001` with just `1` and see how that goes with the tests

marwoodandrew:
@MarkLark86 Tests pass! final build fails, due I think to the version of pip, but can't see where I can change that!

MarkLark86:
You can't, that's on our test infrastructure (using FireQ).

MarkLark86:
The problem is our test server now uses Python v3.10, and I believe a newer version Cerberus is required to support it. Which requires looking at Superdesk & Eve to see if you can update the version to a newer one.

